PR #8, opened 2026-06-10
Fix Codex CLI plugin install instructions in README

### LINKED ISSUE #3: Codex CLI plugin install command in README does not exist
The README shows a Codex CLI install command that does not exist in the current Codex CLI.

### Current README

Under `Agent Skills -> Install -> Codex CLI`, the README says:

```bash
codex plugin add coreai-skills@coreai-models
```

### What happens

With Codex CLI `0.130.0`, `codex plugin add` is not available:

```bash
codex plugin --help
```

returns:

```text
Manage Codex plugins

Usage: codex plugin [OPTIONS] <COMMAND>

Commands:
  marketplace  Manage plugin marketplaces for Codex
  help         Print this message or the help of the given subcommand(s)
```

So a user can follow the README, register the marketplace, then get stuck on the install step.

### Expected

The README should use the current Codex CLI command for installing a marketplace plugin, or call out the exact Codex CLI version needed for this command.

### Why this matters

The skills are one of the main entry points for agent users. If the first install command fails, most people will assume the plugin itself is broken.

### Suggested fix

Update the Codex CLI section with a command that matches the current CLI, or add a short compatibility note like:

```text
These commands require Codex CLI version X.Y.Z or newer.
```

I did not want to guess the intended replacement command here because `codex plugin --help` currently only exposes marketplace management on my install.

srjoglekar246:
Replace the non-existent `codex plugin add coreai-skills@coreai-models` command with the actual interactive flow: launch `codex`, open the plugin manager via `/plugins`, and install `coreai-skills` from the coreai-models marketplace tab.

Fixes #3

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,11 +102,16 @@ Alternatively, register the marketplace from a local git checkout:
 codex plugin marketplace add /path/to/coreai-models
 ```
 
-Install the plugin:
+Launch Codex in your workspace:
 
 ```
-codex plugin add coreai-skills@coreai-models
+codex
 ```
+
+Install the plugin through the interactive browser: once the Codex session is
+active in your terminal, open the plugin manager by typing `/plugins`, locate the
+`coreai-models` marketplace tab (use your arrow keys or the built-in search),
+select `coreai-skills`, and choose Install.
 
 #### Gemini CLI
 


carinapeng (approved):
Thanks!

PR #9, opened 2026-06-10
Tool calling support

carinapeng:
## Purpose

Models like Qwen3 emit tool calls as inline markup mixed into the token stream:

```tool_call
<tool_call>{"name":"get_weather","arguments":{"city":"London"}}</tool_call>
```

Without interception, this markup leaks into `response.content` as plain text and the FoundationModels session never invokes the tool. This PR routes tool call blocks to the correct FM channel events so `LanguageModelSession` tool calling works end-to-end with CoreAI-backed models.

## Changes

**`ToolCallParser.swift`** — new streaming state machine that splits the token stream at configurable open/close markers. Emits `.text` or `.toolCall(id:name:argsJSON:)` events. Handles both Qwen3's `<tool_call>...</tool_call>` tag-pair format and Mistral's `[TOOL_CALLS] [{...}]` single-line array format.

**`CoreAILanguageModel.swift`** — three additions:
- `CoreAIExecutor` detects tool call markers from the tokenizer vocabulary at init and wires `ToolCallParser` into the generation pipeline (`ThinkTagParser` → `ToolCallParser` → channel)
- `transcriptToTokens` now handles `.toolCalls` and `.toolOutput` transcript entries, serializing them for the chat template so multi-turn tool-use history is correctly formatted on each subsequent turn
- `capabilities` is now tokenizer-aware: `.toolCalling` and `.reasoning` are only advertised when the tokenizer's vocabulary contains the corresponding special tokens
## Usage

Standard `LanguageModelSession` with tools 🎉 

```swift
import FoundationModels
import CoreAILanguageModels

struct WeatherTool: Tool {
    let name = "get_weather"
    let description = "Returns the current weather for a city."

    func call(arguments: GeneratedContent) async throws -> String {
        let city = try arguments.value(String.self, forProperty: "city")
        return "72°F and sunny in \(city)."
    }
}

let model = try await CoreAILanguageModel(resourcesAt: modelURL)
let session = LanguageModelSession(model: model, tools: [WeatherTool()])

let response = try await session.respond(
    to: "What's the weather in Cupertino?"
)
// response.content contains the model's final answer after tool execution
print(response.content)
```

### DIFF
--- a/swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
+++ b/swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
@@ -176,7 +176,7 @@ public struct CoreAILanguageModel: LanguageModel {
         /// gpt-oss / Harmony), a different parser is needed; this one
         /// covers the `<open>...</close>` shape.
         private static func detectThinkingMarkers(
-            using tokenizer: some Tokenizer
+            using tokenizer: any Tokenizer
         ) -> (open: String, close: String) {
             let candidates: [(open: String, close: String)] = [
                 ("<think>", "</think>"),
@@ -203,7 +203,7 @@ public struct CoreAILanguageModel: LanguageModel {
         /// matches the bare token without a trailing space — `parseToolCalls`
         /// already trims leading whitespace so optional spacing is handled.
         fileprivate static func detectToolCallMarkers(
-            using tokenizer: some Tokenizer
+            using tokenizer: any Tokenizer
         ) -> (open: String, close: String)? {
             // Standard tag-pair formats — both markers must be special tokens.
             let tagPairs: [(open: String, close: String)] = [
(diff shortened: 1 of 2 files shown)


swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
```diff
@@ -101,61 +104,13 @@ public struct CoreAILanguageModel: LanguageModel {
...
-                CLILogger.log(
-                    "Failed to apply chat template: \(error), falling back to simple encoding",
-                    component: component)
-                let text = messages.compactMap { $0["content"] }.joined(separator: "\n")
-                return tokenizer.encode(text: text)
-            }
-        }
-
-        return nil
+        self.supportsToolCalling =
```
stikves: Let's use `detectToolCallMarkers` here

carinapeng: Good call! Addressed

stikves (approved):

PR #10, opened 2026-06-10
Fix --platform filter being ignored for utility models

### LINKED ISSUE #2: Registry --platform filter is ignored for utility models
The registry lets me ask for iOS utility models, but it still returns a model that is marked as macOS-only.

This is confusing because `--platform` works as a filter for LLMs, and the utility table already has a `platforms` field.

### Repro

From a fresh checkout at `b1cb71b`:

```bash
uv run coreai.model.registry --list-models --type utility --platform iOS
```

The output includes:

```text
depth-anything-3-small   depth   macOS   depth-anything/da3-small   models/depth-anything/export.py
```

This command also succeeds even though the model is macOS-only:

```bash
uv run coreai.model.registry --model-info depth-anything-3-small --type utility --platform iOS --json
```

It returns:

```json
{
  "short_name": "depth-anything-3-small",
  "hf_id": "depth-anything/da3-small",
  "model_type": "depth-anything",
  "task": "depth",
  "export_script": "models/depth-anything/export.py",
  "platforms": [
    "macOS"
  ],
  "notes": null
}
```

### Expected

When I pass `--platform iOS`, utility models should be filtered using the `UtilityModel.platforms` tuple.

So:

```bash
uv run coreai.model.registry --list-models --type utility --platform iOS
```

should not show `depth-anything-3-small`.

And:

```bash
uv run coreai.model.registry --model-info depth-anything-3-small --type utility --platform iOS
```

should probably fail with the same style of error used for LLMs:

```text
Error: model 'depth-anything-3-small' has no --platform 'iOS'
```

### Why this matters

The registry is the first place people will check before exporting a model. If a macOS-only utility model shows up under an iOS filter, users may assume the export path is supported and only find out later.

### Likely fix

Small change in `python/src/coreai_models/model_registry.py`:

- pass `args.platform` into utility filtering
- make `--model-info` for utility models validate `args.platform`
- add a small CLI test so macOS-only utility models do not leak into iOS results

stikves:
The registry CLI accepted --platform for utility model queries but never applied it, causing macOS-only models to appear in iOS-filtered results.

Closes apple/coreai-models#2

### DIFF
--- a/python/src/coreai_models/model_registry.py
+++ b/python/src/coreai_models/model_registry.py
@@ -14,7 +14,7 @@ CLI:
     uv run coreai.model.registry --model-info clip-vit-b32 --type utility --as-export-args
 
 `--type` selects the preset table (`llm`, `diffusion`, or `utility`).
-`--platform` (`macOS` / `iOS`) is meaningful for LLM only.
+`--platform` (`macOS` / `iOS`) filters LLM variants and utility model platforms.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


tjia1818 (approved):

carinapeng (approved):
Thank you!

PR #24, opened 2026-06-11
Use custom AsyncSequences and avoid unstructured concurrency

alejandro-isaza:


stikves:
Thanks

Can you hold on until
https://github.com/apple/coreai-models/pull/16

is merged?

tjia1818 (approved):

swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
```diff
@@ -42,91 +42,25 @@ public struct ConstrainedDecodingStrategy: DecodingStrategy {
...
-                tokenizer: tokenizer,
-                tokenStep: tokenStep
-            )
-            continuation.yield(GenerationResult(text: delta, tokenId: bestToken, rawLogits: logits))
-
-            if session.isTerminated { break }
-        }
-
-        continuation.finish()
+    ) -> any AsyncSequence<GenerationResult, Error> {
```
stikves: To have the control over the streaming generation...

Let's either:

a) Use a concrete type (like the new `InferenceStream`, or another naming)
b) or... Define a new protocol with the side channel

stikves (approved):
Thank you, we should move away from anonymous async sequences.

We can either restructure this onto `InferenceStream` or a similar one with side channel including the new `StopReason`

swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
carinapeng: True, `any AsyncSequence` may be too generic. Rebasing on the InferenceStream from @stikves may solve this?

alejandro-isaza: Fixed

swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
```diff
@@ -237,3 +171,179 @@ public struct ConstrainedDecodingStrategy: DecodingStrategy {
...
+                tokenizer: tokenizer,
+                stopSequences: stopSequences
+            )
+            self.inputTokens =
+                try PromptUtils
+                .maybeApplyTokenizerChatTemplate(input, tokenizer: tokenizer)
+                .map(Int32.init)
+            self.maxTokens = options.maxTokens ?? 512
+
+            try await inferenceEngine.reset()
```
carinapeng: My only clarifying question is whether the move from eager to lazy would introduce any edge cases like a stale engine state between stream creation and the first iteration would be issue.

I maybe be overly cautious here (because we've had problems with cache corruptions when debugging LLMs) but say we have a stale cache where previous call is to write a poem about dogs or something, and then we call Constrained Decoder to give json for a user called John, and model gives json with name "Puppy" -- would this be possible? Like something happens before the next call, and touches the shared engine state so the logits get corrupted?

carinapeng (approved):
Thank you! One comment on now lazily resetting, it in practice will be okay because when we call constrained decoding for instance we iterate immediately

swift/Sources/CoreAILanguageModels/DecodingStrategies/VanillaDecodingStrategy.swift
```diff
@@ -32,172 +34,241 @@ public struct VanillaDecodingStrategy: DecodingStrategy {
...
-                        samplingConfiguration: samplingConfiguration,
-                        options: options,
-                        stopSequences: stopSequences,
-                        with: continuation
-                    )
-                } catch {
-                    continuation.finish(throwing: error)
-                }
+    ) -> VanillaDecodedSequence {
+        VanillaDecodedSequence(
```
stikves: All these would be type-erased by callers. As we don't know the engine type at compile time, and decoding is also dynamic.

But should be a no-op compared to using existential.

swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
stikves: This is a valid concern.

We had a race condition in the FM generation for multi-turn when the work from prior response (which was cut for any reason) continued when the second iteration reset / generate started.

The sequence itself might not be the correct location to handle the exclusion. It might be better to isolate structural and behavioral changes

(In other words, let's not do lazy)

swift/Sources/CoreAILanguageModels/DecodingStrategies/VanillaDecodingStrategy.swift
```diff
@@ -32,172 +34,267 @@ public struct VanillaDecodingStrategy: DecodingStrategy {
...
+                if let matched = stopSequences.matchedSequence(recentTokens: recentTokens) {
+                    CLILogger.log("✅ Stop sequence detected at tokens: \(recentTokens)", level: 2)
+                    decodingSpan.end()
+                    stopped = true
+
+                    // Record why the engine stream ended before abandoning it.
+                    let matchedText = tokenizer.decode(tokens: matched.map(Int.init))
+                    engineSequence?.setStopReason(.stopSequence(matchedText))
+                    engineSequence = nil
+                    upstreamIterator = nil
```
stikves, line 228: Specifically here we need to eagerly stop the upstream iterator. As waiting for ARC would not clean up GPU tasks in time

(Tested the race condition with a local copy of this branch and a small stress script)

swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
alejandro-isaza: Fixed!

swift/Sources/CoreAILanguageModels/DecodingStrategies/VanillaDecodingStrategy.swift
alejandro-isaza: Added a drain loop.

PR #29, opened 2026-06-12
Dynamic and Batch Support for Object Detector

kevchengcodes:
## Summary

Enables the `ObjectDetector` to run dynamic models as well as static models with `batch_size > 1`.

Changes include resolving dynamic dimensions similar to how `llm-runner` does it, constructing a batched input tensor based on the list of input images, and post-processing each output.

Models can have dynamic batch size, image input height/width. Note that batched images will be resized to a common size when constructing the (B,C,H,W) input tensor. Users can optionally input a custom input height/width for their dynamic batch.

## Verification
Tested with YoloS:
```bash
uv run models/yolo/export.py --dynamic
swift run -c release object-detector --model exports/yolos-base_float32_dynamic.aimodel --image img.png --output-image out.png
```
New unit tests to validate batch support is working.

### DIFF
--- a/swift/Sources/CoreAIObjectDetector/ObjectDetector.swift
+++ b/swift/Sources/CoreAIObjectDetector/ObjectDetector.swift
@@ -129,10 +129,9 @@ public struct ObjectDetector {
     ///    parameters. Batch is always `images.count`. Dynamic spatial dims
     ///    are filled from `parameters.inputHeight` / `inputWidth` (which
     ///    have struct-level defaults).
-    /// 2. Preprocess each image sequentially into a `[3, H, W]` Float buffer.
-    /// 3. Concatenate the per-image buffers into the `[B, 3, H, W]` input
-    ///    NDArray and run a single forward pass.
-    /// 4. Slice each batch slot from the outputs and decode independently,
+    /// 2. Allocate the `[B, 3, H, W]` input NDArray and preprocess each
+    ///    image directly into its batch slot, then run a single forward pass.
+    /// 3. Slice each batch slot from the outputs and decode independently,
     ///    returning `images.count` detection lists in input order.
     public func detect(images: [CGImage], parameters: DetectionParameters) async throws
         -> [[DetectedObject]]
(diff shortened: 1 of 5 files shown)


swift/Sources/CoreAIObjectDetector/ObjectDetector.swift
```diff
@@ -87,68 +87,200 @@ public struct ObjectDetector {
                 "No array descriptor for image input '\(imageInputName)'"
             )
         }
-        let imageArray = NDArray(descriptor: imageDescriptor)
-        _ = try await function.run(inputs: [imageInputName: imageArray])
+        let defaults = DetectionParameters()
+        let warmupShape = zip(imageDescriptor.shape, [1, 3, defaults.inputHeight, defaults.inputWidth])
```
carinapeng: Here we warmup with default, not the shape the caller will actually use at inference time. We should check if this is expected -- we might be getting kernels compiled for the default shapes but inferences runs at a different shape which would be a different kernel, so we'd be wasting cost here

Maybe `warmup()` should accept the parameters you detect (`DetectionParameters`, or a `BatchPlan`?) so we compile for the right shape

swift/Sources/CoreAIObjectDetector/ObjectDetector.swift
```diff
@@ -87,68 +87,200 @@ public struct ObjectDetector {
...
             )
         }
-        let height = expectedShape[2]
-        let width = expectedShape[3]
-        let floatPixels = try ImagePreprocessor(
-            targetSize: CGSize(width: width, height: height),
+
+        let plan = try Self.planBatch(
+            expectedShape: expectedShape,
+            imageSizes: images.map { CGSize(width: $0.width, height: $0.height) },
```
carinapeng: I am wondering if we use the width and height here? We seem to use .count for the batching, but H and W always come from expectedShape in static or parameters.inputHeight/inputWidth in dynamic

If you think this is expected, we should change function signature :)

If not, maybe we should use them?

swift/Sources/CoreAIObjectDetector/ObjectDetector.swift
kevchengcodes: Updated to use the same shape during warmup

swift/Sources/CoreAIObjectDetector/ObjectDetector.swift
kevchengcodes: This is expected, so I changed the function signature to show we're just using count

carinapeng (approved):
Thank you!

PR #30, opened 2026-06-12
ci: run CI on self-hosted Apple Silicon macOS runners

DawerG:
## Summary
Moves all CI jobs onto self-hosted Apple Silicon macOS runners (`[self-hosted, macos, tahoe, ARM64]`).

The previous `ubuntu-latest` `python-test` job could not work: `coreai-torch`/`coreai-core` ship as Apple-Silicon macOS wheels and the model-unit tests need MPS. Lint and swift-format are co-located on the same runners.

## Changes
- `lint`, `swift-format`, `python-test` → `runs-on: [self-hosted, macos, tahoe, ARM64]`
- Swift toolchain now comes from Xcode via `DEVELOPER_DIR` (drops the Linux-oriented `swift-actions/setup-swift`; uses `swift format lint`)
- Pin `actions/checkout` to a full commit SHA and install `uv` via its official install script instead of third-party setup actions
- `if: github.repository == 'apple/coreai-models'` guard so forks don't run on the self-hosted runners
- Added `timeout-minutes` to every job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,26 +12,37 @@ concurrency:
 
 jobs:
   lint:
-    runs-on: ubuntu-latest
+    if: github.repository == 'apple/coreai-models'
+    runs-on: [self-hosted, macos, tahoe, ARM64]
+    timeout-minutes: 15
     steps:
-      - uses: actions/checkout@v4
-      - uses: astral-sh/setup-uv@v4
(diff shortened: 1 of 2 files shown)


tjia1818 (approved):

carinapeng (approved):
Thanks for this and thanks for the format change!

stikves (approved):

PR #32, opened 2026-06-12
Add cancel() and isBusy to InferenceEngine protocol

stikves:
Adds lifecycle management APIs to the InferenceEngine protocol so callers can gracefully cancel in-flight generation and query busy state. 

All three engine implementations (Sequential, Pipelined, StaticShape) and MockEngine are updated with concrete implementations. 

CoreAILanguageModel now calls cancel() before reset() to ensure clean state transitions.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
@@ -41,6 +41,12 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
     private let engineInUse = Atomic<Bool>(false)
     let config: ModelConfig
 
+    // Generation lifecycle
+    private let _activeToken = Mutex<GenerationToken?>(nil)
+    private let _generationTask = Mutex<Task<Void, Never>?>(nil)
+
+    var isBusy: Bool { _activeToken.withLock { $0 != nil } }
+
     init(
         config: ModelConfig,
         preparedModel: PreparedModel,
@@ -104,9 +110,20 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
         let stopReasonStore = StopReasonStore()
         let (base, outputContinuation) =
             AsyncThrowingStream<InferenceOutput, any Error>.makeStream()
-        Task {
+
+        let token = GenerationToken()
+        _activeToken.withLock { $0 = token }
+
+        let task = Task {
             self.acquireEngine()
-            defer { self.releaseEngine() }
+            defer {
+                self.releaseEngine()
+                // Only clear if this generation still owns both slots
+                if self._activeToken.withLock({ $0 === token }) {
+                    self._activeToken.withLock { $0 = nil }
+                    self._generationTask.withLock { $0 = nil }
+                }
+            }
             do {
                 let (tokenStream, tokenContinuation) =
                     AsyncThrowingStream<InferenceEngine.TokenId, any Error>.makeStream()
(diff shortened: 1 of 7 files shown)


swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -139,23 +140,24 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
...
-    /// Wait for any in-flight generate() Task to return the engine.
-    private func drain() {
-        var attempts = 0
-        while engineInUse.load(ordering: .acquiring) {
-            attempts += 1
-            if attempts > 5000 {
-                fatalError("Engine not returned after drain() — tokenSequence Task stuck?")
-            }
-            Thread.sleep(forTimeInterval: 0.001)
+    var isBusy: Bool { engineInUse.load(ordering: .acquiring) }
```
kevchengcodes: nit: this var is hidden here, could we move this up to the other defined attributes? same for the other engines.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -347,20 +348,25 @@ public final class CoreAISequentialEngine: InferenceEngine, @unchecked Sendable
 
     // MARK: - Lifecycle
 
-    /// Wait for any in-flight generate() Task to finish.
-    private func drain() {
+    public var isBusy: Bool { generating.withLock { $0 } }
```
kevchengcodes: any reason why some engines don't specify public but some do?

stikves: this depends on whether the engine itself is public or not. that is another concern for a possible follow up, though

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -347,20 +348,25 @@ public final class CoreAISequentialEngine: InferenceEngine, @unchecked Sendable
...
+
+    public func cancel() async throws {
```
carinapeng: My understanding is that cancel() in static engine has a slightly different meaning than in pipelined, and sequential shares that same definition as static here

Pipelined engine cancels a real background Task so cancel actually does stop the work. but the static and sequential engines has no background Task so generation runs inside the consumer's next() calls, so generating only becomes false when next() runs and sees cancel being requested.

So cancel here kind of means: we ask the consumer to stop if it keeps pulling. Should we restructure the static/sequential path so cancellation doesn't depend on consumer making progress? 

The main problem I see here is that cancelRequested is only checked inside next(), so semantically the idea of cancel is different between pipelined and the other engines

carinapeng: To make more concrete if a consumer is paused and not calling next() the check never runts so cancel() in that case means throwing a timeout

stikves: Yes that is a valid concern. The iterator and the engine are a bit decoupled, and it might make sense to cancel on the engine side, and have the iterator be invalid.

stikves: Added a Token concept. And only the iterator that contains the token would talk to the engine

tjia1818 (approved):

PR #33, opened 2026-06-12
Align diffusion bundles with metadata.json v0.2 schema

stikves:
Diffusion exports now produce metadata.json (same v0.2 schema as LLM and segmenter bundles) instead of the legacy pipeline.json format.

The pipeline config lives under a "diffusion" key in metadata.json, alongside standard fields (kind, name, assets, source, compression, compilation) shared by all bundle types.

Swift runner: reads metadata.json first. If only pipeline.json is found, prints a deprecation error to stderr and asks for a clean export, the message will be removed in upcoming weeks.


python/src/coreai_models/diffusion/pipeline.py
```diff
@@ -296,33 +298,63 @@ def _save_tokenizer(model_id: str, output_path: Path, hf_pipe: Any, overwrite: b
 
 
 # ---------------------------------------------------------------------------
-# pipeline.json
+# metadata.json (v0.2 schema — aligned with LLM and segmenter bundles)
```
kevchengcodes:
```suggestion
# metadata.json
```

python/src/coreai_models/diffusion/pipeline.py
```diff
@@ -296,33 +298,63 @@ def _save_tokenizer(model_id: str, output_path: Path, hf_pipe: Any, overwrite: b
...
+        "metadata_version": METADATA_VERSION,
+        "kind": "diffusion",
+        "name": output_path.name,
+        "assets": assets,
+        "diffusion": diffusion_config,
+        "source": {
+            "model_definition": "torch",
+            "hf_model_id": model_id,
+        },
+        "compression": compression if compression != "none" else None,
```
kevchengcodes, line 338: why not just `"compression": compression`? We don't want "none" as a possible output?

stikves: Yes, "none" as a verbatim string will not increase the information here, but will increase parsing difficulty.

kevchengcodes (approved):

swift/Sources/CoreAIDiffusionPipeline/Pipelines/PipelineDescriptor.swift
```diff
@@ -105,18 +105,27 @@ public struct PipelineDescriptor: Codable, Sendable {
...
     /// by inspecting the actual model descriptors.
     public static func resolve(at url: URL, config: ConfigSource = .auto) throws -> PipelineDescriptor {
         switch config {
         case .auto:
-            let configURL = url.appendingPathComponent("pipeline.json")
-            if FileManager.default.fileExists(atPath: configURL.path) {
-                return try load(from: configURL)
+            let metadataURL = url.appendingPathComponent("metadata.json")
+            if FileManager.default.fileExists(atPath: metadataURL.path) {
+                return try loadFromMetadata(at: metadataURL)
```
carinapeng, line 120: Just to clarify, would this break for any assets we have been supporting? Ie. Flux where we have extra keys? Should we throw an error or validate somewhere for the key mapping in the assets?

carinapeng: Wanna make sure we are not silently dropping Flux assets like transformer_512 for  instance. In general it would be helpful to have runtime validation as we go through different iterations of assets and runtime checks

carinapeng (approved):
Happy to merge this, but some considerations for more validation as we go through different iterations of assets, would prefer to have it in this PR or a follow up

swift/Sources/CoreAIDiffusionPipeline/Pipelines/PipelineDescriptor.swift
stikves: The half/tiled variants (Transformer_512, VAEDecoder_half) are resolved by filename scanning in Flux2Pipeline+Resources.swift

This is going to be another follow up to clean on the Flux side. This PR is concerned more about the metadata.

PR #34, opened 2026-06-12
Fix Whisper export to support autoregressive decoding

carinapeng:
## Purpose 
The existing export traced decoder_input_ids at length 1 with no dynamic shapes, producing a model specialised to a single token. At inference time this means each decode step has no context — giving empty transcript

This fix adds dynamic_shapes for decoder_input_ids and traces with the 4-token forced prefix so torch.export generalises the sequence dimension to allow full autoregressive decoding.

### DIFF
--- a/models/whisper/export.py
+++ b/models/whisper/export.py
@@ -107,8 +107,17 @@ def create_whisper(
 
     example_inputs = reference_inputs(model_name, dtype)
 
+    example_inputs["decoder_input_ids"] = torch.tensor(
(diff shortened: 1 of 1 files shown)


models/whisper/export.py
stikves, line 47: I think it might be better to fix it here, rather than patching later

stikves (approved):

models/whisper/export.py
carinapeng: We have an authored one that is much more optimized, I can look into it as a part of that as well in the export script

PR #51, opened 2026-06-16
Add reset(to:), processedTokenCount, and implicit prefix caching

stikves:
Extends the InferenceEngine protocol with partial KV cache reset and automatic prefix detection:

- reset(to: tokenIndex): truncate KV cache to keep first N positions
- processedTokenCount: read-only, tracks engine's current cache position
- TokenHistory: shared utility for memcmp-based prefix detection
- Pipelined engine: implicit prefix caching — engine auto-detects common prefix between consecutive generate() calls, only processes new tokens
- Sequential/Static: implicit rewind when input is shorter than cache (full divergence triggers zero-fill, prefix extension rewinds counter)

All engines pass explicit reset parity test. Pipelined passes implicit prefix caching test (extend + diverge). Sequential/Static pass by wiring explicit reset.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
@@ -38,6 +38,8 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
     typealias ConfigType = ModelConfig
 
     nonisolated(unsafe) private var engine: EngineImpl
+    nonisolated(unsafe) private var history = TokenHistory()
+    nonisolated(unsafe) private(set) var lastPrefixHitCount: Int = 0
     private let engineInUse = Atomic<Bool>(false)
     let config: ModelConfig
 
@@ -46,6 +48,8 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
     private let _generationTask = Mutex<Task<Void, Never>?>(nil)
 
     var isBusy: Bool { _activeToken.withLock { $0 != nil } }
+
+    var processedTokenCount: Int { engine.processedTokenCount }
 
     init(
         config: ModelConfig,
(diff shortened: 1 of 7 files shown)


swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -128,18 +132,46 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
                 let (tokenStream, tokenContinuation) =
                     AsyncThrowingStream<InferenceEngine.TokenId, any Error>.makeStream()
 
+                // Implicit prefix caching: resolve input against history
+                var (commonPrefix, resolvedNewTokens) = self.history.resolve(input: input)
```
carinapeng, line 136: It seems that resolve runs once for pipelined engine, at generate() setup inside the Task, for static we resolve on every token, is that expected? (Seems like we don't need to do that after step 0 because history only grows by tokens it just added?)

Sequential resolves at the start, why is the condition that's gating it different from others?

Can we make the implementations for resolving shared among three engines to avoid error proneness?

stikves: Yes, that is a bug in the static one, let's fix it now.
Pipelined has a different condition, but sequential and static can match.

As for sharing more code, it would require a larger refactor.

carinapeng: Thanks for the fix for static engine! (gating to step==0 and rewinding to commonPrefix looks good)

Should we rewind the sequential one to commonPrefix as well? (seems to still rewind to `inputTokens.count-1`, is that expected? Let's make sure all edge cases are tested because silently having wrong logits from infra changes is tricky to manage 

These are great infra change proposals, we should definitely consider refactors, but they can be error prone

stikves: Yes, it should also use prefix match. Updated

alejandro-isaza (approved):

carinapeng (approved):
Thank you!

PR #52, opened 2026-06-17
Dump GPU Optests

YifanShenSZ:
Dump op tests and run on non-CPU backends, similar to [coreai-torch](https://github.com/apple/coreai-torch/pull/19)

Usage: In addition to our usual pytest
```shell
uv run pytest python/tests/ -p no:warnings -n 4
```
Add 2 optional args
1. `--compute-unit-kind=gpu`
2. `--dump-optests`

Testing: Locally tested with
```shell
uv run pytest python/tests/ -p no:warnings -n 4 --compute-unit-kind=gpu --dump-optests
```

### DIFF
--- a/python/tests/_runner_infra/run/runners/coreai_runner.py
+++ b/python/tests/_runner_infra/run/runners/coreai_runner.py
@@ -9,21 +9,15 @@ import asyncio
 import shutil
 from contextlib import AsyncExitStack
 from pathlib import Path
-from typing import TYPE_CHECKING
 
 import torch
+from coreai.authoring import AIProgram
+from coreai.runtime import InferenceFunction, NDArray
 from typing_extensions import Self, override
 
 from ...common.types.dependency_types import Tensor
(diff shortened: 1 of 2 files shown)


python/tests/_runner_infra/run/runners/coreai_runner.py
```diff
@@ -9,29 +9,25 @@
...
-    from coreai.runtime import InferenceFunction, NDArray
-
 
 class CoreaiRuntime:
     async def _async_load_model(
         self: Self,
         coreai_program: AIProgram,
         asset_path: Path,
     ) -> InferenceFunction:
+        from tests.conftest import get_test_specialization_options
```
stikves, line 29: I don't think we should wire test code into production.

Can we try either:

A - Adding an optional "AssetDumper" structure that we pass here
B - Using decorations / patches at the test level to induce this behavior?

tjia1818:
stale for too long, closing for now. Feel free to re-open if needed.

PR #54, opened 2026-06-17
Add speech-runner: Swift inference for Whisper model

carinapeng:
## Purpose
Swift CLI that loads a CoreAI Whisper export and transcribes audio

## Changes
  - `swift/Sources/Tools/speech-runner/SpeechRunnerMain.swift` — loads either export format, greedy decodes with forced prefix + KV cache, decodes tokens to text via bundled tokenizer
  - `swift/Sources/Tools/speech-runner/WhisperMel.swift` — mel spectrogram computation in Swift
  - `Package.swift` — registers speech-runner target
  
  ## Testing
  Usage
  `swift run speech-runner <model-path> <audio.flac>`
  
  Tested running both converted model from ToT main and reauthored model from branch `carina/whisper`. No impact on model performance or export quality in this PR


stikves (approved):

swift/Sources/CoreAISpeech/SpeechModel.swift
```diff
@@ -0,0 +1,130 @@
...
+    private var encOutShape: [Int]?
+
+    /// Load a model from a bundle directory.
+    ///
+    /// - Parameters:
+    ///   - bundleURL: Directory containing encoder.aimodel and decoder.aimodel.
+    ///   - decoder: Decode strategy. Defaults to ``WhisperDecoder``.
+    ///   - melConfig: Mel spectrogram parameters. Defaults to ``MelConfig/whisper``.
+    public init(
+        bundleURL: URL,
```
kevchengcodes: nit: most of the other modalities use `resourcesAt` in the initializer

swift/Sources/CoreAISpeech/SpeechBundle.swift
```diff
@@ -0,0 +1,125 @@
...
+            let folderName = "models--" + name.replacingOccurrences(of: "/", with: "--")
+            let snapshotsDir = cacheRoot.appending(path: "\(folderName)/snapshots")
+            if let snapshot = try? FileManager.default.contentsOfDirectory(
+                atPath: snapshotsDir.path
+            ).first {
+                return try? await AutoTokenizer.from(
+                    modelFolder: snapshotsDir.appending(path: snapshot))
+            }
+        }
+        return nil
```
kevchengcodes, line 68: what happens when nil is returned?

swift/Sources/CoreAISpeech/SpeechModel.swift
carinapeng: Good catch, will align w other modalities!

swift/Sources/CoreAISpeech/SpeechBundle.swift
carinapeng: We'd throw a `SpeechError.missingTokenizer`. The tokenizer is optional during bundle loading -- the files are valid and loadable without it (the encoder and decoder files), we only need it at the detokenize step

kevchengcodes: Makes sense! It could also make sense to throw here instead of waiting to throw later, but as long as we catch the behavior somehow.

kevchengcodes (approved):

PR #65, opened 2026-06-25
Add VLM inference infrastructure: engine, protocol, and CLI support

stikves:
Runtime:
- MultimodalInferenceEngine protocol with encodeImage() and generate()
- CoreAISequentialVLMEngine: vision encoder + projector + embed_tokens + LLM decoder with scatter-merge of image embeddings at placeholder positions
- EmbeddedInput type wrapping NDArray embeddings with position metadata
- VisionConfig in LanguageConfig for image_size, patch_size, token count/id
- LanguageBundle parses top-level "vision" block from metadata.json

CLI (llm-runner):
- --image flag routes through VLM engine when bundle kind is .vlm
- Chat template detection with generic fallback for prompt construction
- Accumulated token decode for correct spacing
- Stop sequence support in VLM path

Supports any VLM that exports 3 components (vision.aimodel, embed.aimodel, model.aimodel) with a vision config block in metadata.json. Model-family- specific export code lives in internal/python.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
+++ b/swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
@@ -22,18 +22,23 @@ public struct LanguageConfig: Codable, Sendable, Equatable {
     /// known role conventions (`main`, `extend_<N>`, `load_embeddings`, ...).
     public let functionMap: FunctionMap?
 
+    /// Vision-specific configuration. Nil for text-only language models.
+    public let vision: VisionConfig?
+
     public init(
         tokenizer: String,
         vocabSize: Int,
         maxContextLength: Int,
         embeddedTokenizer: Bool = true,
-        functionMap: FunctionMap? = nil
+        functionMap: FunctionMap? = nil,
+        vision: VisionConfig? = nil
     ) {
         self.tokenizer = tokenizer
         self.vocabSize = vocabSize
         self.maxContextLength = maxContextLength
         self.embeddedTokenizer = embeddedTokenizer
         self.functionMap = functionMap
+        self.vision = vision
     }
 
     enum CodingKeys: String, CodingKey {
@@ -42,6 +47,7 @@ public struct LanguageConfig: Codable, Sendable, Equatable {
         case maxContextLength = "max_context_length"
         case embeddedTokenizer = "embedded_tokenizer"
         case functionMap = "function_map"
+        case vision
     }
 
     public init(from decoder: Swift.Decoder) throws {
@@ -51,6 +57,7 @@ public struct LanguageConfig: Codable, Sendable, Equatable {
         self.maxContextLength = try c.decode(Int.self, forKey: .maxContextLength)
         self.embeddedTokenizer = try c.decodeIfPresent(Bool.self, forKey: .embeddedTokenizer) ?? true
         self.functionMap = try c.decodeIfPresent(FunctionMap.self, forKey: .functionMap)
+        self.vision = try c.decodeIfPresent(VisionConfig.self, forKey: .vision)
     }
 
     // MARK: - Additional Stop Tokens
@@ -135,5 +142,40 @@ public struct LanguageConfig: Codable, Sendable, Equatable {
         }
 
         return Array(result)
+    }
+}
+
+/// Vision-specific configuration for VLM bundles.
+/// Nil for text-only language models.
+public struct VisionConfig: Codable, Sendable, Equatable {
+    /// Input image size (square). Vision encoder expects this resolution.
+    public let imageSize: Int
+
+    /// Patch size for the vision transformer.
+    public let patchSize: Int
+
+    /// Number of embedding tokens produced per image after projection.
+    public let imageTokenCount: Int
+
+    /// Token ID used as a placeholder in the text sequence for image positions.
+    public let imageTokenId: Int32
+
+    public init(
+        imageSize: Int,
+        patchSize: Int,
+        imageTokenCount: Int,
+        imageTokenId: Int32
+    ) {
+        self.imageSize = imageSize
+        self.patchSize = patchSize
+        self.imageTokenCount = imageTokenCount
+        self.imageTokenId = imageTokenId
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case imageSize = "image_size"
+        case patchSize = "patch_size"
+        case imageTokenCount = "image_token_count"
+        case imageTokenId = "image_token_id"
     }
 }
--- a/swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
+++ b/swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
@@ -160,16 +160,35 @@ public struct VisionConfig: Codable, Sendable, Equatable {
     /// Token ID used as a placeholder in the text sequence for image positions.
     public let imageTokenId: Int32
 
+    /// Per-channel normalization mean (RGB). Defaults to CLIP values when omitted.
+    public let imageMean: [Double]
+
+    /// Per-channel normalization std (RGB). Defaults to CLIP values when omitted.
+    public let imageStd: [Double]
+
+    /// Pixel rescale factor applied before normalization. Defaults to 1.0 when omitted.
+    public let rescaleFactor: Double
+
+    /// CLIP normalization (Qwen VL, Pixtral, InternVL, Phi-3.5-vision).
+    public static let clipMean = [0.48145466, 0.4578275, 0.40821073]
+    public static let clipStd = [0.26862954, 0.26130258, 0.27577711]
+
     public init(
         imageSize: Int,
         patchSize: Int,
         imageTokenCount: Int,
-        imageTokenId: Int32
+        imageTokenId: Int32,
+        imageMean: [Double]? = nil,
+        imageStd: [Double]? = nil,
+        rescaleFactor: Double? = nil
     ) {
         self.imageSize = imageSize
         self.patchSize = patchSize
         self.imageTokenCount = imageTokenCount
         self.imageTokenId = imageTokenId
+        self.imageMean = imageMean ?? Self.clipMean
+        self.imageStd = imageStd ?? Self.clipStd
+        self.rescaleFactor = rescaleFactor ?? 1.0
     }
 
     enum CodingKeys: String, CodingKey {
@@ -177,5 +196,19 @@ public struct VisionConfig: Codable, Sendable, Equatable {
         case patchSize = "patch_size"
         case imageTokenCount = "image_token_count"
         case imageTokenId = "image_token_id"
+        case imageMean = "image_mean"
+        case imageStd = "image_std"
+        case rescaleFactor = "rescale_factor"
+    }
+
+    public init(from decoder: Swift.Decoder) throws {
+        let c = try decoder.container(keyedBy: CodingKeys.self)
+        self.imageSize = try c.decode(Int.self, forKey: .imageSize)
+        self.patchSize = try c.decode(Int.self, forKey: .patchSize)
+        self.imageTokenCount = try c.decode(Int.self, forKey: .imageTokenCount)
+        self.imageTokenId = try c.decode(Int32.self, forKey: .imageTokenId)
+        self.imageMean = try c.decodeIfPresent([Double].self, forKey: .imageMean) ?? Self.clipMean
+        self.imageStd = try c.decodeIfPresent([Double].self, forKey: .imageStd) ?? Self.clipStd
+        self.rescaleFactor = try c.decodeIfPresent(Double.self, forKey: .rescaleFactor) ?? 1.0
     }
 }
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
@@ -264,9 +264,9 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
             throw InferenceRuntimeError.invalidOutputType(
                 "Cannot get descriptor for '\(logitsName)'")
         }
-        guard logitsDesc.scalarType == .float16 else {
+        guard logitsDesc.scalarType == .float16 || logitsDesc.scalarType == .bfloat16 else {
             throw InferenceRuntimeError.unsupportedLogitsType(
-                "Only float16 logits supported, got \(logitsDesc.scalarType)")
+                "Only float16/bfloat16 logits supported, got \(logitsDesc.scalarType)")
         }
         self.logitsDescriptor = logitsDesc
 
@@ -381,7 +381,7 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
 
         return EmbeddedInput(
             embeddings: projectedEmbeddings,
-            imageTokenPositions: placeholderRange
+            embeddingPositions: placeholderRange
         )
     }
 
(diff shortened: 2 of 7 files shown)


swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1116 @@
...
+
+        // Load LLM inference function
+        guard let llmFn = try llmModel.model.loadFunction(named: config.function) else {
+            throw InferenceRuntimeError.genericError(
+                "Cannot load function '\(config.function)'")
+        }
+        self.llmFunction = llmFn
+
+        // Select image preprocessor based on vision config
+        // Default to gemma3/SigLIP preset; can be extended for other encoders
```
carinapeng: This hardcoding would break silently for any non ImageNet encoders, for Qwen3 VL we use CLIP for instance

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1116 @@
...
+        self.imagePreprocessor = ImagePreprocessor(
+            targetSize: CGSize(
+                width: config.visionConfig.imageSize,
+                height: config.visionConfig.imageSize
+            ),
+            mean: (0.485, 0.456, 0.406),
+            std: (0.229, 0.224, 0.225),
+            rescaleFactor: 1.0
+        )
```
carinapeng, line 323: I like that we are abstracting to the metadata (VisionConfig), so we can have something like deriving the CLIP mean/std in metadata from the different encoders?

and then pass those into the ImageProcessor

swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
```diff
@@ -137,3 +144,38 @@ public struct LanguageConfig: Codable, Sendable, Equatable {
         return Array(result)
     }
 }
+
+/// Vision-specific configuration for VLM bundles.
+/// Nil for text-only language models.
+public struct VisionConfig: Codable, Sendable, Equatable {
```
carinapeng, line 150: See earlier comment, should we add normalization fields to the VisionConfig? If ImageNet doesn't need them we can make them optional

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1116 @@
...
+                "scatterMerge: image embeddings have \(imgSeqLen) tokens, need \(imageTokenCount)")
+        }
+
+        // Validate all positions are within bounds
+        guard let maxPos = imagePositions.last, maxPos < seqLen else {
+            throw InferenceRuntimeError.invalidArgument(
+                "scatterMerge: image position \(imagePositions.last ?? -1) exceeds sequence length \(seqLen)")
+        }
+
+        imageEmbeddings.view(as: Float16.self).withUnsafePointer { imgPtr, _, _ in
```
carinapeng, line 577: I think we consistently promise bf16 in addition to f16, but here we are hardcoding. This means a bf16 embed/vision export would silently break

For Qwen3 BL we have f16, but let's at least add a guard of the scalar type :) and maybe change some the other comments to be consistent

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1116 @@
...
+// MARK: - Core AI Sequential VLM Engine
+
+/// Sequential inference engine for Vision-Language Models using Core AI APIs.
+///
+/// ## Model Contract
+///
+/// Manages three model functions (potentially from separate `.aimodel` bundles):
+///
+/// 1. **Vision encoder** (`encode_image`):
+///    - Input: `pixel_values` (Float32, shape `[1, 3, H, W]`)
```
carinapeng, line 49: Is this a contract we want to enforce?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1116 @@
...
+
+        // --- LLM pipeline ---
+
+        guard let llmDesc = llmModel.model.functionDescriptor(for: config.function) else {
+            throw InferenceRuntimeError.functionNotFound(config.function)
+        }
+        self.llmFunctionDescriptor = llmDesc
+
+        // Validate LLM architecture: expects inputs for embeddings + position_ids, states for KV cache
+        guard llmDesc.inputNames.count == 2 else {
```
carinapeng, line 228: This is right call!

Though I am wondering if we can have a better way of contracting with export, if there is a model that do not conform with this spec, in which case this might seem too brittle of a count check

But non-blocking, we can keep this in mind for iterating on this

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: Make sense, adding more configs for alternative model structures

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: This seems to be the most common case, but yes, it make sense to add a config

swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
stikves: done

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: We can start here, and update as more models arrive.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: Yes, it is a compromise between hard coding input names, or expecting a certain structured order. Maybe we can add overrides in the config in future.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: Yes, bf16 support should arrive. However when I did a quick check, none of the VLMs we might support in short run has bf16 (only some community patches on existing fp16 ones)

I'll leave a TODO and come back later. As we might promise this support prematurely if we cannot test it.

swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
```diff
@@ -137,3 +144,71 @@ public struct LanguageConfig: Codable, Sendable, Equatable {
...
+    public let patchSize: Int
+
+    /// Number of embedding tokens produced per image after projection.
+    public let imageTokenCount: Int
+
+    /// Token ID used as a placeholder in the text sequence for image positions.
+    public let imageTokenId: Int32
+
+    /// Per-channel normalization mean (RGB). Defaults to CLIP values when omitted.
+    public let imageMean: [Double]
```
kevchengcodes, line 164: fine as-is, but the `CoreAIShared` uses `(CGFloat, CGFloat, CGFloat)` for mean and std, you could consider making this the same

swift/Sources/CoreAILanguageModels/InferenceEngines/InferenceEngine.swift
```diff
@@ -288,6 +288,38 @@ public enum InferenceRuntimeError: Error, LocalizedError {
...
+/// Engine that supports vision/audio input in addition to text tokens.
+///
+/// The typical flow:
+/// 1. `encodeImage(at:)` — preprocess + run vision encoder, return embeddings
+/// 2. `generate(with: EmbeddedInput, ...)` — scatter-merge embeddings into
+///    token sequence and run prefill + decode
+///
+/// The caller owns the embeddings and decides caching strategy.
+public protocol MultimodalInferenceEngine: InferenceEngine {
+    /// Encode an image into embeddings suitable for injection into the LLM.
```
kevchengcodes: should this comment say VLM not LLM?

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
```diff
@@ -748,6 +802,166 @@ struct LLMRunner: AsyncParsableCommand, Sendable {
...
+            print("Error: image not found at \(imagePath)")
+            throw ExitCode.failure
+        }
+
+        guard let visionConfig = bundle.visionConfig else {
+            print("Error: VLM bundle missing 'vision' config in metadata.json")
+            throw ExitCode.failure
+        }
+
+        if !CLILogger.isVerbose {
```
kevchengcodes, line 833: why the NOT on this condition?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1120 @@
...
+///
+/// ## Inference Flow
+///
+/// 1. `encodeImage(at:)` — preprocess image, run vision encoder + projector, return `EmbeddedInput`
+/// 2. `generate(with: EmbeddedInput, tokens:, ...)` — embed tokens, scatter-merge with vision
+///    embeddings at placeholder positions, run LLM prefill, then standard autoregressive decode
+///
+/// KV cache is managed identically to `CoreAISequentialEngine`: starts small and grows
+/// dynamically with 2x expansion.
+public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchecked Sendable {
```
kevchengcodes, line 73: can we remove the unchecked?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1120 @@
...
+            featuresDesc.shape.map { $0 < 0 ? 1 : $0 })
+        var featuresArray = NDArray(descriptor: resolvedFeaturesDesc)
+
+        // Execute encode_image
+        var outputViews = InferenceFunction.MutableViews()
+        outputViews.insert(&featuresArray, for: featuresOutputName)
+
+        _ = try await visionFunction.run(
+            inputs: [pixelInputName: pixelArray],
+            outputViews: consume outputViews
```
kevchengcodes, line 430: I think I've seen the compiler complain about this being unsafe, if you remove the unchecked from Sendable. Ignore if this isn't the case

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1120 @@
...
+        }
+    }
+
+    public func cleanup() {
+        let cleanupSpan = InstrumentsProfiler.beginCleanup(engine: "CoreAIVLM")
+        CLILogger.log("CoreAI VLM engine cleanup complete")
+        cleanupSpan.end()
+    }
+
+    public func warmup(queryLength: Int, sampling: SamplingConfiguration?) async throws {
```
kevchengcodes: neither of these args are used, let's remove them

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1120 @@
...
+
+    private func ensureKVCapacity(forContextLength needed: Int) throws {
+        guard needed > currentKVCapacity else { return }
+        guard needed <= config.maxContextLength else {
+            throw InferenceRuntimeError.invalidState(
+                "Context length \(needed) exceeds maximum \(config.maxContextLength)")
+        }
+
+        var newCapacity = currentKVCapacity
+        while newCapacity < needed { newCapacity *= 2 }
```
kevchengcodes, line 868: what if `newCapacity` starts at 0? we'll enter an infinite loop

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -0,0 +1,1120 @@
...
+                    dstPtr.advanced(by: dstOff).update(
+                        from: srcPtr.advanced(by: srcOff), count: copySize)
+                }
+            }
+        }
+    }
+
+    // MARK: - Helpers
+
+    private func zeroFill(_ array: inout NDArray) {
```
kevchengcodes, line 920: is there a more efficient way to do this?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
carinapeng: This is a protocol requirement so we can't just remove the args I believe
But this is a good catch, maybe we can prefix _

carinapeng:
https://github.com/apple/coreai-models/pull/68 export of Qwen3 VL verifies export of runner with the latest changes, thanks for addressing the comments! @stikves

carinapeng (approved):

swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
stikves: Let's leave this as TODO if we later move to CoreAIShared

swift/Sources/CoreAILanguageModels/InferenceEngines/InferenceEngine.swift
stikves: Good catch

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
stikves: `--verbose` enables bunch of other diagnostic messages, this is a fallback for the plain output

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: added a floor of 1, good catch

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: Yes, but this matches existing pattern. Maybe we can do a follow up.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: We have mutable state, which should be safe with our patterns, but requires deeper analysis.

Should keep this as a reminder

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: 3 of our 4 engines currently use `@uncheched Sendable`, let's do a follow up.

This is currently safe, as it was tested with real data. However, yes, we should do better.

swift/Sources/CoreAILanguageModels/Bundle/LanguageConfig.swift
kevchengcodes: Looks like you already do use CoreAIShared and just cast to CGFloats later

Issue #70, opened 2026-06-30
Populate performance summary in VLM runner / engine


carinapeng:
### What are you trying to build?

Currently the text only models support this, but we should add this to the VLM path as well, ie the token counts and timings. In `runVLMInference`, call `setPromptTokenCount(vlmTokens.count)` and wrap the prefill + generation loop

Another design is to instrument `CoreAISequentialVLMEngine` to record `.prompt/.extend` spans the way the text engines do, so metrics work for any caller

### Where are the current docs or utilities unclear?

N/A

### Expected improvement

More readable and informative outputs

carinapeng:
https://github.com/apple/coreai-models/pull/72 implements this

PR #72, opened 2026-07-01
VLM: wire performance instrumentation and logits output

stikves:
Wire performance metrics and logits output for the VLM inference path, matching the LLM path behavior.

## Changes

- Report prompt throughput (prefill t/s) and generation throughput in the performance summary
- Wire `--print-logits` for VLM: shows top-5 token probabilities per generated step
- Wire `--save-logits` for VLM: saves top-K logits to JSON (same format as LLM path)
- Make `TokenLogits` and `TopLogitEntry` properties public (cross-module access)

## Sample output

```
$ llm-runner --model <vlm-bundle> --image photo.jpg --prompt "What is this?" --max-tokens 10

Generating...
The image features a small kitten sitting on

Performance Summary:
==================================================
Model Load: 15299.1ms
Prompt:     1028.3ms, 590 tokens, 573.8 tokens/sec
Generation: 480.8ms, 9 tokens, 18.7 tokens/sec
Total:      24.479s
==================================================
```

With `--print-logits`:
```
Generating...
  logits top5: [450]=26.812 [910]=25.406 [512]=24.422 [319]=22.609 [739]=22.109 The
  logits top5: [1967]=25.562 [9088]=19.609 [7623]=18.797 [15373]=18.734 [17739]=17.516 image
  ...
```

With `--save-logits /tmp/vlm_logits.json`:
```json
{
  "tokens": [{
    "token_id": 450,
    "incremental_text": "The",
    "top_logits": [
      {"token_id": 450, "incremental_text": "The", "logit": 26.8125},
      {"token_id": 910, "incremental_text": "This", "logit": 25.406}
    ]
  }]
}
```

## Test plan

- [x] Build passes
- [x] VLM shows correct prompt/generation t/s in performance summary
- [x] `--print-logits` displays top-5 logits per token during VLM generation
- [x] `--save-logits` produces valid JSON with top-K entries
- [x] Verbose output (`--verbose`) shows full timing breakdown table
- [ ] Verify no regression on text-only LLM path


swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
```diff
@@ -886,27 +886,62 @@ struct LLMRunner: AsyncParsableCommand, Sendable {
...
+        await PerformanceMetrics.shared.setPromptTokenCount(vlmTokens.count)
+
         let tokenStream = try vlmEngine.generate(
             with: embeddedInput,
             tokens: vlmTokens,
             samplingConfiguration: samplingConfiguration,
-            inferenceOptions: InferenceOptions(maxTokens: maxTokens)
+            inferenceOptions: InferenceOptions(
+                maxTokens: maxTokens,
+                includeLogits: printLogits || saveLogits != nil
```
carinapeng, line 900: Can we double check on how we are hooking this up with --save-logits? For LLMs we consume this https://github.com/stikves/coreai-models/blob/20cfdc5e34b066709b5cc67dc2133514d8dd8612/swift/Sources/Tools/llm-runner/LLMRunnerMain.swift#L640

But we don't seem to do the same for VLM?

carinapeng: I think right now we are computing the logits, but are we saving them? :)

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
```diff
@@ -886,27 +886,62 @@ struct LLMRunner: AsyncParsableCommand, Sendable {
...
         for try await output in tokenStream {
+            if promptSpan != nil {
+                promptSpan?.end()
+                promptSpan = nil
+                extendSpan = InstrumentsProfiler.beginExtend(step: 0, tokens: 1)
+            }
+
             let token = output.tokenId
+            if printLogits {
+                print("\n  raw token=\(token)", terminator: "")
```
carinapeng: Why are we printing these? Printing every raw token (like including EOS?) Is this intended or was for debugging?

carinapeng:
Thank you for taking this on @stikves !

We seem to have implemented the runner-level design ( in runVLMInference, call setPromptTokenCount(vlmTokens.count) and wrap the prefill + generation) https://github.com/apple/coreai-models/issues/70#:~:text=In%20runVLMInference%2C%20call,+%20generation%20loop 

I proposed engine level here as well, seems to me it could be a more sustainable design because if we instrument `CoreAISequentialVLMEngine` then metrics can work for any caller, that's how we do it for text engines as well

I wonder if that'd be a better design to be more generic?

tjia1818 (approved):

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
stikves: Good catch.

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
stikves: Yes, debug artifacts... removing...

PR #79, opened 2026-07-02
VLM: fix EmbeddedInput shape handling, error recovery, and parallel loading

stikves:
## Summary

- **EmbeddedInput.tokenCount**: correctly handle 2D `[seq_len, hidden_dim]` tensors (was returning `hidden_dim` instead of `seq_len`). Now a stored property computed once at init.
- **EmbeddedInput.seqLen(of:)**: centralized helper for seq_len extraction — scatterMerge and other call sites use one canonical path instead of inline shape checks.
- **scatterMerge**: replace `precondition(float16)` with a thrown error for bfloat16 model compatibility.
- **LLMRunnerMain**: add comment explaining sequential PreparedModel.prepare (parallel caused runtime errors).
- **Removed** premature `--max-tiles` flag (tiling model not implemented yet).

## Test plan

- [x] Verify tokenCount returns correct seq_len for both 2D and 3D embeddings
- [x] Confirm bfloat16 model gets a catchable error instead of a crash
- [x] VLM inference still works end-to-end

### DIFF
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
@@ -385,7 +385,7 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
 
         CLILogger.log("VLM encodeImage complete: \(tokenCount) embedding tokens")
 
-        return EmbeddedInput(
+        return try EmbeddedInput(
             embeddings: projectedEmbeddings,
             embeddingPositions: placeholderRange
         )
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
@@ -560,8 +560,8 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
                     + "expected \(imageTokenCount) from config. Check prompt template.")
         }
 
-        let seqLen = EmbeddedInput.seqLen(of: textEmbeddings)
-        let imgSeqLen = EmbeddedInput.seqLen(of: imageEmbeddings)
+        let seqLen = textEmbeddings.shape[1]
+        let imgSeqLen = imageEmbeddings.shape[1]
         guard imgSeqLen >= imageTokenCount else {
             throw InferenceRuntimeError.invalidArgument(
                 "scatterMerge: image embeddings have \(imgSeqLen) tokens, need \(imageTokenCount)")
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
@@ -574,10 +574,10 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
         }
 
         // Copy image embeddings into placeholder positions.
-        precondition(
-            imageEmbeddings.scalarType == .float16,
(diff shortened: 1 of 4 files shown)


swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
```diff
@@ -374,6 +374,7 @@ struct LLMRunner: AsyncParsableCommand, Sendable {
             )
             let vlmConfig = VLMModelConfig(base: baseConfig, visionConfig: visionConfig)
 
+            // Sequential to avoid runtime errors with concurrent model preparation.
```
kevchengcodes, line 374: this doesn't match the PR description:

> LLMRunnerMain: parallelize 3 independent PreparedModel.prepare calls with async let for faster VLM startup

swift/Sources/CoreAILanguageModels/InferenceEngines/EmbeddedInput.swift
kevchengcodes: > EmbeddedInput.init: add precondition validating tokenCount == embeddingPositions.count to catch mismatches early

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
```diff
@@ -180,7 +180,9 @@ struct LLMRunner: AsyncParsableCommand, Sendable {
     var imagePath: String?
 
     @Option(
+        name: .customLong("max-tiles"),
```
carinapeng: Where is maxTiles read? It's defined here but where do we use it?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -570,10 +570,10 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
         }
 
         // Copy image embeddings into placeholder positions.
-        precondition(
```
carinapeng, line 577: Good call to convert this to an invalid input type throw!

swift/Sources/CoreAILanguageModels/InferenceEngines/EmbeddedInput.swift
```diff
@@ -26,7 +26,11 @@ public struct EmbeddedInput: Sendable {
 
     /// Number of embedding tokens (seq_len dimension).
     public var tokenCount: Int {
-        embeddings.shape.count >= 2 ? embeddings.shape[1] : 0
+        switch embeddings.shape.count {
+        case 3...: embeddings.shape[1]  // [batch, seq_len, hidden_dim]
```
carinapeng: We are adding these conditionals, can we verify that the 2d embeddings never actually flow into scatter merge? Or do we normalize the layout in one place? 

We should be consistent

carinapeng (commented):
This looks good to me overall but before approving I think a few things should be fixed, one is on how we read maxTiles, and the other is consistency - seems correct for the 2d case but scatter merge still issues 3D: [1, seq, hidden]

And let's do the testing on the Qwen3 VL as well since that is released :)

swift/Sources/Tools/llm-runner/LLMRunnerMain.swift
stikves: Ah... this should be other way around. The tiling model is not ready yet.

carinapeng:
Should we rebase this PR?

Also an alternative to the max tiling is to restore the missing property here in `LLMRunnerMain.swift`, we'd get a build failure right now on ToT main which breaks running VLMs

```
      183 +        name: .customLong("max-tiles"),
      184          help: "Maximum tiles for image splitting (overrides model config). 1 = single crop, no tiling.")
      185 +    var maxTiles: Int?
```

aolivares85-collab (approved):

kevchengcodes (approved):

carinapeng (approved):
LGTM, let's merge after verifying commits!

tjia1818 (approved):

PR #80, opened 2026-07-02
Make generate() async to serialize back-to-back turns

stikves:
## Summary

Fixes a crash when reusing an engine for rapid multi-turn conversations:

```
CoreAIPipelinedEngine.swift:82: Fatal error: Trying to acquire engine when it's already in use
```

The `InferenceEngine.generate()` method is now `async throws` (was `throws`). The pipelined engine awaits the prior generation's Task before starting a new one, preventing the fatalError on rapid back-to-back calls.

## Root cause

PR #64 removed the per-turn `engine.reset()` call to enable multi-turn KV cache reuse. That `reset()` also served as the serialization point between consecutive turns (it called `drain()` internally). Without it, a new `generate()` call can race with the prior generation's GPU pipeline drain.

## Fix

The pipelined engine now cancels and awaits any in-flight generation at the top of `generate()`:

```swift
if let priorTask = _generationTask.withLock({ $0 }) {
    _activeToken.withLock { $0?.cancel() }
    await priorTask.value
}
```

This preserves KV cache state — prefix caching handles reuse automatically across turns.

## Test plan

- [ ] Existing `CancelAPITests` pass
- [ ] Existing `UnifiedGenerationAPITests` pass (prefix caching still works)
- [ ] New: back-to-back generate() calls do not crash
- [ ] New: rapid-fire 10-turn stress test (zero delay between turns)
- [ ] New: generate() after cancel() works cleanly
- [ ] Manual: multi-turn LanguageModelSession with Qwen3-0.6B pipelined engine

### DIFF
--- a/swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
+++ b/swift/Sources/CoreAILanguageModels/DecodingStrategies/ConstrainedDecodingStrategy.swift
@@ -122,7 +122,7 @@ public struct ConstrainedDecodingStrategy: DecodingStrategy {
         constrainedOptions: InferenceOptions
     ) async throws -> (Int32?, [LogitsScalarType]?) {
         var rawLogits: [LogitsScalarType]? = nil
-        for try await output in try inferenceEngine.generate(
+        for try await output in try await inferenceEngine.generate(
             with: inputTokens,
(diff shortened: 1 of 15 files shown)


carinapeng:
Thanks for looking into this! The root cause is helpful tracing :)

tjia1818 (approved):

PR #81, opened 2026-07-02
Default `max-context-length` of `4096` for LLM iOS platform exports

Lewis300:
- Default `max-context-length` of `4096` for LLM iOS platform exports **including experimental models not included in the model registry**
- Add check to ensure that the max context length in the `ExportConfig` is less than or equal to that of the HF config
- Add tests to ensure this new infra functions as expected

### DIFF
--- a/python/src/coreai_models/export/pipeline.py
+++ b/python/src/coreai_models/export/pipeline.py
@@ -25,6 +25,7 @@ from coreai_opt.palettization.config.palettization_config import KMeansPalettize
 from transformers import AutoConfig, AutoTokenizer
 
 from coreai_models.export._constants import (
+    IOS_DEFAULT_MAX_CONTEXT_LENGTH,
     QUANT_TRACE_OFFSET,
     QUANT_TRACE_QUERY_LEN,
     TRACE_KV_CACHE_SEQ_LEN,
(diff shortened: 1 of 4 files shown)


tjia1818 (approved):

python/src/coreai_models/export/pipeline.py
```diff
@@ -151,7 +152,27 @@ async def _async_export_model(config: ExportConfig) -> str:
     # ---- 2. Load model ----
     target_dtype = _resolve_precision(config.compute_precision)
 
+    # The model's native context window from its HuggingFace config. Any
+    # user-provided override must not exceed it.
+    native_max_ctx = getattr(hf_config, "max_position_embeddings", None)
```
stikves, line 157: there is too much nesting here, we can probably make it shorter like

```python
  native_max_ctx = getattr(hf_config, "max_position_embeddings", None)

  max_context_length = config.max_context_length
  if max_context_length is None and config.variant == "iOS":
      max_context_length = min(IOS_DEFAULT_MAX_CONTEXT_LENGTH, native_max_ctx or IOS_DEFAULT_MAX_CONTEXT_LENGTH)

  if max_context_length is not None and native_max_ctx is not None and max_context_length > native_max_ctx:
      raise ValueError(
          f"--max-context-length ({max_context_length}) exceeds the model's "
          f"max_position_embeddings ({native_max_ctx}). Choose a value <= {native_max_ctx}."
      )

```

Lewis300: Done

stikves (approved):

PR #85, opened 2026-07-06
Avoid copying embedding table in static shape runner

blevine1:
We currently provide a pre-allocated output view for the embedding table to CoreAI, which has it copy the embedding table into it. However since it's a constant, if we don't use the pre-allocated output view flow, it'll return an NDArray backed by the constant directly without making a copy.


carinapeng (approved):

blevine1:
- [ ] I'm investigating a potential perf issue being observed with this change. Please do not merge it

tjia1818:
Closing (too long since open). Pls reopen once ready

PR #36, opened 2026-06-13
Fix Gemma stop tokens: read additional EOS from tokenizer config

stikves:
This fixes #25

The root cause is upstream not exposing the required fields, and eventually we should try to have them fix this at the origin. However we can also manually read the tokenizer additional information.

```bash
% swift run -c release llm-runner --model exports/gemma_3_4b_it_4bit_dynamic --max-tokens 250
Building for production...
[1 / 1]

Build complete! (1.10 sec)

⏳ Preparing AI asset from source... done in 0.571s

Generating...
I’m doing well, thank you for asking! As a large language model, I don’t really *feel* in the same way humans do, but my systems are running smoothly and I’m ready to chat. 😊

How are *you* doing today? Is there anything you'd like to talk about or any tasks you'd like me to help with?

⏱️  Performance Summary:
==================================================
Model Load: 571.2ms
Prompt:     84.9ms, 15 tokens, 176.6 tokens/sec
Generation: 1873.1ms, 79 tokens, 42.2 tokens/sec
Total:      8.984s
==================================================
```

PR #91, opened 2026-07-08
Lazy runner design: defer engine load

carinapeng:
## Purpose
Introduces lazy inference-engine lifecycle management for the FoundationModels adoption layer. Instead of CoreAILanguageModel holding a live engine from construction, engine load/unload is now owned by a new ModelResources type: lazy by default, loaded on first respond(...) (or explicitly via load()), borrowed per-generation, and releasable via unload(). This lets callers construct a model cheaply, query its on-disk footprint, and control when the (expensive) engine actually loads.

## Changes

- **ModelResources (new)** — owns the engine's load/unload lifecycle. Concurrent callers share a single in-flight load; `withEngine { }` borrows the engine for the duration of a generation and defers teardown if an `unload()` races an active borrow. Shared per-configuration via a weak registry.
- **CoreAILanguageModel** — now builds directly from LanguageBundle (removes the `CoreAIRunner.makeLanguageModel()` hop). Adds:
  - `LoadMode { .lazy, .eager }` (default `.lazy`; `.eager` loads tokenizer + engine concurrently)
  - `load()` / `unload()` / `estimatedSizeOnDiskBytes`
  - guided-generation capability gating that works before the engine is resident (`isGuidedGenerationSupported`)
  - reasoning models default to a higher max-token budget
- **prewarm** — now kicks off the engine load in the background (non-blocking) instead of a synchronous semaphore wait.
- **ResourceManaging** — moved from CoreAIDiffusionPipeline/Components/ to CoreAIShared/Runtime/; diffusion pipelines updated to import it from the shared module.
- **FileSize (new)** — `URL.recursiveFileSizeInBytes()` helper backing `estimatedSizeOnDiskBytes`.
- **Tests** — ModelResourcesTests (load/unload/borrow races via an injectable loader), FileSizeTests, and PublicInterfaceTests additions.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
+++ b/swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
@@ -14,30 +14,43 @@ import Tokenizers
 /// Wraps any `InferenceEngine` (pipelined, sequential, or static-shape) and exposes it
 /// through the FoundationModels `LanguageModel` protocol. It uses the modern `tokenSequence()`
 /// API for efficient streaming token generation.
-///
 /// ## Engine Selection
 /// The engine type is determined by `EngineFactory` based on model structure:
-/// - **Pipelined**: GPU-accelerated with pipeline-depth-matched buffering (fastest for GPU models)
+/// - **Pipelined**: GPU-accelerated with double buffering (fastest for GPU models)
 /// - **Sequential**: CPU-based synchronous execution (fallback)
 /// - **Static-shape**: Neural Engine optimized for chunked static models
 ///
 /// ## Usage
 /// ```swift
-/// let engine = try await EngineFactory.createEngine(...)
-/// let model = CoreAILanguageModel(engine: engine, tokenizer: tokenizer)
+/// let model = try await CoreAILanguageModel(resourcesAt: url)  // .lazy by default
+/// print(model.estimatedSizeOnDiskBytes ?? 0)
+/// try await model.load()                                       // optional; respond auto-loads
 /// let session = LanguageModelSession(model: model)
+/// // ... generate ...
+/// model.unload()
 /// ```
 public struct CoreAILanguageModel: LanguageModel {
+    public enum LoadMode: Sendable {
+        case lazy
+        case eager
+    }
+
     // MARK: - Properties
 
-    private let engine: any InferenceEngine
-    private let tokenizer: any Tokenizer
-    private let modelIdentifier: String
-    private let samplingConfig: SamplingConfiguration
-    private let vocabSize: Int?
+    private let url: URL
+    private let variant: String?
+    private let kvCacheStrategy: KVCacheStrategy
+    fileprivate let samplingConfig: SamplingConfiguration
+    fileprivate let bundle: LanguageBundle
+    fileprivate let tokenizer: any Tokenizer
+    fileprivate let thinkingMarkers: (open: String, close: String)
+    fileprivate let toolCallMarkers: (open: String, close: String)?
     private let supportsToolCalling: Bool
-    private let supportsReasoning: Bool
-    private let additionalEosTokenIds: [Int32]
+    fileprivate let supportsReasoning: Bool
+    fileprivate let resources: ModelResources
+    /// All EOS-like token IDs beyond the tokenizer's main `eosTokenId` — e.g.
+    /// Gemma's `<end_of_turn>`, read from tokenizer_config.json at init.
+    fileprivate let additionalEosTokenIds: [Int32]
 
     // MARK: - Protocol Requirements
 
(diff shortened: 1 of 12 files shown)


swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
```diff
@@ -302,47 +299,75 @@ public struct CoreAILanguageModel: LanguageModel {
...
-                )
+            let effectiveSamplingConfig = makeSamplingConfig(
+                from: request.generationOptions, base: model.samplingConfig)
+            let defaultMaxTokens = model.supportsReasoning ? 2048 : 512
+            let maxTokens = request.generationOptions.maximumResponseTokens ?? defaultMaxTokens
+
+            // Borrow the engine for the whole generation.
+            try await resources.withEngine { engine in
+                // Reset engine state for new generation
+                try await engine.reset()
```
kevchengcodes: does resetting here conflict with #64 ?

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAIRunner.swift
```diff
@@ -69,36 +68,6 @@ public struct CoreAIRunner {
         )
     }
 
-    /// Creates a LanguageModel for FM API usage.
-    func makeLanguageModel() async throws -> CoreAILanguageModel {
```
kevchengcodes, line 73: Looks like yes, but can you confirm all the functionality we need from this model is reproduced? Just want to make sure #36 functionality is retained

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
carinapeng: Good catch! Removed

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
```diff
@@ -285,7 +282,7 @@ public struct CoreAILanguageModel: LanguageModel {
             let tokenizationSpan = InstrumentsProfiler.beginTokenization(inputLength: 0)
             let promptTokens = Self.makeTokens(
                 from: Array(request.transcript),
-                using: tokenizer,
+                using: model.tokenizer,
```
kevchengcodes, line 285: CoreAILanguageModel has a file private tokenizer attribute. Can we use that or is this different?

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
```diff
@@ -240,38 +264,11 @@ public struct CoreAILanguageModel: LanguageModel {
...
-
-            semaphore.wait()
+        // MARK: - Prewarm (FoundationModels, synchronous)
 
-            if let error: any Error = warmupError.withLock(\.self) {
-                throw error
-            }
+        /// Kicks off the engine load in the background.
+        public func prewarm(model: CoreAILanguageModel, transcript: Transcript) {
+            Task { try? await resources.loadResources() }
```
kevchengcodes, line 271: don't need to do the engine warmup anymore?

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
carinapeng: Btw another reconciliation is making the caller of `generate()` and `decode()` to be `try await` instead of `try` because we changed those funcs to be `async`

And we are keeping the extra EOS tokens feature as well

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
```diff
@@ -14,30 +14,43 @@ import Tokenizers
 /// Wraps any `InferenceEngine` (pipelined, sequential, or static-shape) and exposes it
 /// through the FoundationModels `LanguageModel` protocol. It uses the modern `tokenSequence()`
 /// API for efficient streaming token generation.
-///
 /// ## Engine Selection
 /// The engine type is determined by `EngineFactory` based on model structure:
-/// - **Pipelined**: GPU-accelerated with pipeline-depth-matched buffering (fastest for GPU models)
+/// - **Pipelined**: GPU-accelerated with double buffering (fastest for GPU models)
```
kevchengcodes: this reverts a change from #53

carinapeng: Updated!

carinapeng:
CoreAILanguageModel.swift syncing with recent PRs
  - #24 try await strategy.decode → present
  - 02a8edd Gemma additional-EOS → reconciled (additionalEosTokenIds)
  - #53 → doc comment reconciled
  - #64 remove engine.reset() → reconciled
  - #80 try await engine.generate → present

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAIRunner.swift
carinapeng: Yep yep it is retained

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
carinapeng: Warmup still happens, it's, moved to `ModelResources.loadEngine`

swift/Sources/CoreAILanguageModels/LanguageModel/CoreAILanguageModel.swift
carinapeng: It's the same thing essentially, model.tokenizer is that `fileprivate attribute`
tokenizer lives on CoreAILanguageModel and not on the executor by the design

tjia1818 (approved):

PR #114, opened 2026-07-17
Add GPU-based constrained sampling for pipelined engine

stikves:
Summary

  - GPU-accelerated grammar-constrained generation for the pipelined engine
  - Applies xgrammar bitmasks directly in the MPSGraph sampler, eliminating per-token logit transfer to CPU
  - ConstrainedGenerationCapable protocol for capability-based routing
  - Session cache with checkout/checkin pattern for xgrammar reuse across calls

  ## Performance (Qwen3-4B, relative to unconstrained pipelined baseline)

  | Configuration | Relative | Notes |
  |---|---|---|
  | Pipelined unconstrained | 100% | roofline |
  | Sequential constrained | 75% | existing option |
  | **Pipelined constrained + jump-forward** | **82%** | new option, tested with long string responses  |

  ## Changes

  - **`ConstrainedGenerationCapable`** — Protocol for engines that support GPU-side bitmask application
  - **`ConstrainedSessionHandle`** — Encapsulated handle with narrow API, stores tokenizer for JF
  - **`PipelinedConstrainedDecodingStrategy`** — AsyncSequence-based strategy with single-use guard, early-stop on consumer drop
  - **`MPSGraphSamplers`** — Bitmask expansion graph, `applyBitmask` on encode paths
  - **`ConstrainedGenerationSession`** — `fillBitmask(into:)` with `BitmaskResult` enum, `rollback`, `findJumpForwardString`
  - **C bridge extensions** — `rollback`, `findJumpForwardString`, `isCompleted`
  - **`llm-runner`** — Routes to pipelined strategy, performance metrics for constrained path
  - **Integration tests** — Mock engine exercising cache reuse, schema invalidation, stop sequences, error paths

 ## Test plan
  
  - [x] Existing unit tests pass (sampler, session, pipeline gate)
  - [x] New GPU sampler tests: bitmask blocks dominant, all-ones matches unconstrained, only-allowed tokens
  - [x] Integration tests via MockConstrainedEngine (no Metal required)
  - [x] swift-format clean
  - [x] End-to-end on-device validation (greedy outputs match sequential path)

### DIFF
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
@@ -9,6 +9,7 @@ import Foundation
 import Metal
 import MetalPerformanceShaders
 import Synchronization
+import Tokenizers
 import os
 
 // MARK: - Timing
@@ -41,7 +42,7 @@ private let minimumMPSNDArrayBufferSize = 64
 /// - Pipeline-depth-matched buffer rotation for CPU/GPU overlap
 /// - Growing KV cache with pipelined expansion
 /// - All tensors are owned MTLBuffers — Core AI never allocates/frees them
-final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
+final class CoreAIPipelinedEngine: InferenceEngine, ConstrainedGenerationCapable, Sendable {
     typealias ConfigType = ModelConfig
 
     nonisolated(unsafe) private var engine: EngineImpl
@@ -59,6 +60,9 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
     // Generation lifecycle
     private let _activeToken = Mutex<GenerationToken?>(nil)
     private let _generationTask = Mutex<Task<Void, Never>?>(nil)
+
+    // Constrained session cache — checkout/checkin pattern for reuse across calls
+    private let _constrainedSessionCache = Mutex<ConstrainedSessionHandle?>(nil)
 
     var isBusy: Bool { _activeToken.withLock { $0 != nil } }
 
@@ -242,6 +246,64 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
         return GenerationSequence(base: base, stopReasonStore: stopReasonStore)
     }
 
+    // MARK: - Constrained Generation (GPU bitmask path)
+
+    /// Generate tokens with grammar-constrained sampling on GPU.
+    ///
+    /// Unlike `generate()`, this method applies an xgrammar bitmask inside the GPU sampler
+    /// each step, enabling constrained decoding without falling back to the sequential engine.
+    ///
+    /// The loop is semi-pipelined: inference overlaps with bitmask computation, but sampling
+    /// waits per token (grammar state is inherently sequential).
+    func generateConstrained(
+        with input: [TokenId],
+        samplingConfiguration: SamplingConfiguration,
+        maxTokens: Int,
+        session: ConstrainedSessionHandle
+    ) throws -> AsyncThrowingStream<TokenId, Error> {
+        let (stream, continuation) = AsyncThrowingStream<TokenId, Error>.makeStream()
+
+        let sessionBox = session
+        let isCancelled = Atomic<Bool>(false)
+        continuation.onTermination = { _ in
+            isCancelled.store(true, ordering: .relaxed)
+        }
+
+        let token = GenerationToken()
+        _activeToken.withLock { $0 = token }
+
+        let task = Task {
+            self.acquireEngine()
+            defer {
+                self.releaseEngine()
+                self.storeConstrainedSessionForReuse(sessionBox)
+                if self._activeToken.withLock({ $0 === token }) {
+                    self._activeToken.withLock { $0 = nil }
+                    self._generationTask.withLock { $0 = nil }
+                }
+            }
+            self.engine.reset()
+            do {
+                try await self.engine.runConstrainedCompletion(
+                    prompt: input,
+                    sampler: samplingConfiguration,
+                    sessionBox: sessionBox,
+                    maxTokens: maxTokens,
+                    isCancelled: isCancelled,
+                    yieldingTo: continuation
+                )
+                continuation.finish()
+            } catch is CancellationError {
+                continuation.finish()
+            } catch {
+                continuation.finish(throwing: error)
+            }
+        }
+        _generationTask.withLock { $0 = task }
+
+        return stream
+    }
+
     /// Wait for any in-flight generate() Task to return the engine.
     private func drain() {
         var attempts = 0
@@ -325,6 +387,98 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
         acquireEngine()
         defer { releaseEngine() }
         try await engine.performWarmup(queryLength: queryLength, samplingConfig: sampling)
+    }
+}
+
+// MARK: - Constrained Session Cache
+
+extension CoreAIPipelinedEngine {
+    /// Check out a constrained session from the cache, or create a new one.
+    /// The cache slot is emptied — concurrent calls get independent sessions.
+    func getOrCreateConstrainedSession(
+        jsonSchema: String,
+        tokenizer: any Tokenizer,
+        vocabSize: Int,
+        stopTokenIds: [Int32]?
+    ) throws -> ConstrainedSessionHandle {
+        if let cached: ConstrainedSessionHandle = _constrainedSessionCache.withLock({ cache in
+            guard let box = cache, box.schema == jsonSchema else {
+                cache = nil
+                return nil
+            }
+            cache = nil
+            return box
+        }) {
+            cached.reset()
+            return cached
+        }
+        let session = try ConstrainedGenerationSession(
+            jsonSchema: jsonSchema,
+            tokenizer: tokenizer,
+            vocabSize: vocabSize,
+            stopTokenIds: stopTokenIds
+        )
+        return ConstrainedSessionHandle(session: session, tokenizer: tokenizer)
+    }
+
+    /// Return a session box to the cache for reuse by the next call.
+    func storeConstrainedSessionForReuse(_ box: ConstrainedSessionHandle) {
+        _constrainedSessionCache.withLock { $0 = box }
+    }
+}
+
+// MARK: - Constrained Session Handle
+
+/// Thread-safe handle to a `ConstrainedGenerationSession`.
+///
+/// Access is serialized by the engine's `acquireEngine()`/`releaseEngine()` atomic
+/// gate — only one Task mutates the session at a time. The `@unchecked Sendable`
+/// conformance is justified by this single-writer guarantee.
+///
+/// The handle exposes only the operations the engine loop needs — no raw
+/// session access from outside.
+package final class ConstrainedSessionHandle: @unchecked Sendable {
+    /// Schema string is immutable after init — cached to avoid repeated access.
+    let schema: String
+
+    /// Tokenizer used for jump-forward string encoding.
+    private let tokenizer: any Tokenizer
+
+    private var _session: ConstrainedGenerationSession
+
+    init(session: consuming ConstrainedGenerationSession, tokenizer: any Tokenizer) {
+        self.schema = session.schema
+        self.tokenizer = tokenizer
+        self._session = session
+    }
+
+    var isTerminated: Bool { _session.isTerminated }
+
+    func acceptToken(_ token: Int32) -> Bool {
+        _session.acceptToken(token)
+    }
+
+    func fillBitmask(into ptr: UnsafeMutablePointer<Int32>) -> ConstrainedGenerationSession.BitmaskResult {
+        _session.fillBitmask(into: ptr)
+    }
+
+    func reset() { _session.reset() }
+
+    func rollback(_ numTokens: Int = 1) { _session.rollback(numTokens) }
+
+    func findJumpForwardString() -> String? {
+        _session.findJumpForwardString()
+    }
+
+    /// Decode a sequence of token IDs back into a string using the stored tokenizer.
+    func decodeTokens(_ tokens: [Int32]) -> String {
+        tokenizer.decode(tokens: tokens.map(Int.init))
+    }
+
+    /// Tokenize a string for jump-forward injection. Uses the stored tokenizer
+    /// without adding special tokens.
+    func tokenizeForJumpForward(_ string: String) -> [Int32] {
+        tokenizer.encode(text: string).map(Int32.init)
     }
 }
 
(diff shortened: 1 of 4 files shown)


swift/Sources/CoreAILanguageModels/InferenceEngines/ConstrainedGenerationCapable.swift
```diff
@@ -0,0 +1,43 @@
...
+    ) throws -> ConstrainedSessionHandle
+
+    /// Stream constrained token generation using GPU-side bitmask application.
+    ///
+    /// The engine resets state, prefills the prompt (unconstrained), then enters
+    /// a semi-pipelined loop: inference overlaps bitmask computation, but sampling
+    /// waits per token (grammar state is inherently sequential).
+    ///
+    /// The handle is returned to the engine's cache automatically when the Task completes.
+    func generateConstrained(
```
carinapeng, line 41: Are we draining an already in flight generation in `generateConstrained`?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -1004,6 +1158,341 @@ private struct EngineImpl: ~Copyable {
...
+        var valState = unsafe InferenceFunction.AsyncMutableValue(
+            unsafeBuffer: valBuffer, byteOffset: 0,
+            scalarType: valueCacheScalarType, shape: valShape, strides: valStrides
+        )
+
+        var asyncStates = InferenceFunction.AsyncMutableViews()
+        asyncStates.insert(&keyState, for: keyCacheName)
+        asyncStates.insert(&valState, for: valueCacheName)
+
+        let logitsBuffer = logits.metalBuffer
```
carinapeng, line 1442: So this is a single shared logits buffer. With one buffer the step n+1's inference would overwrite step n's logits before the sampler reads them. This might be why it is not fully working as pipelined. We should investigate further

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -1004,6 +1158,341 @@ private struct EngineImpl: ~Copyable {
...
+        sampler: SamplingConfiguration,
+        sessionBox: ConstrainedSessionHandle,
+        maxTokens: Int,
+        isCancelled: borrowing Atomic<Bool>,
+        yieldingTo continuation: AsyncThrowingStream<Int32, Error>.Continuation
+    ) async throws {
+        let gpuSampler = try getOrCreateSampler(for: sampler)
+
+        // Get the bitmask buffer from whichever concrete sampler we have
+        let bitmaskBuffer: MTLBuffer
```
carinapeng, line 1188: Are we using one bit mask buffer per sampler? If we fill in place each step, illing the bit mask for n+1 may be writing the same memory for GPU that is still reading for n.

swift/Sources/CoreAILanguageModels/InferenceEngines/ConstrainedGenerationCapable.swift
stikves: Yes -- the drain happens in the public generate() path (lines 131-136) which cancel+awaits any prior _generationTask before the strategy calls generateConstrained. So in practice the engine slot is always free by the time we enter here.

That said, generateConstrained is a protocol method -- a direct caller could skip the drain. I will add a precondition assert (or inline drain) inside the method body to make it self-contained.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: Good catch on the general principle. In this constrained path specifically, it is safe: runConstrainedCompletion awaits each token synchronously before encoding the next step (the comment at line 1165 explains: grammar state is sequential). So step n logits are fully consumed by the sampler before step n+1 inference writes to the same buffer.

Added a clarifying comment at the binding site. If we ever relax this to true overlap, we would need double-buffered logits -- but for grammar-constrained generation, sequential is inherent (each bitmask depends on the prior token grammar state).

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: Same story as the logits buffer -- the constrained loop is fully sequential: we fill the bitmask on CPU, encode inference+sampling, then await the token result before filling the next bitmask. The GPU has finished reading the bitmask buffer by the time we write the next mask.

The semi-pipelined aspect is only that inference overlaps bitmask computation (we compute the next mask while the prior inference is in flight), but we never fill the buffer until after the prior step sample completes. So one bitmask buffer per sampler is safe here.

carinapeng (commented):
Thanks for looking into this. Getting this working is tricky because constrained decoding has some data dependency that vanilla decoding doesn't. From the grammar side, the accepted token to fill bit masking runs on CPU and the bit mask for token n+1 would depend on the sampled value of token n so this GPU-CPU roundtrip might need more design change. Happy to chat more on this and this work would probably benefit from breaking this into smaller pieces!

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: In other words, this is a "hybrid" solution, and only does proper pipelining when we are not contained in the grammar (fast forward tokens)

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: Again the primary benefits are:

- GPU masks
- Batch pipelining for fast forward

Depending on the model and the grammar, this closes some or in cases all the gap between pipelined and sequential engines.

stikves:
> Thanks for looking into this. Getting this working is tricky because constrained decoding has some data dependency that vanilla decoding doesn't. From the grammar side, the accepted token to fill bit masking runs on CPU and the bit mask for token n+1 would depend on the sampled value of token n so this GPU-CPU roundtrip might need more design change. Happy to chat more on this and this work would probably benefit from breaking this into smaller pieces!

Masking happens on the GPU sampler:
https://github.com/apple/coreai-models/pull/114/changes#diff-26ae392b0ec2c8ff20901ac369417462a4f015174091b838203b599c9992db7dR239

Yes, there is a CPU work for the xgrammar, however unlike the sequential engine, it does not "stop world" to generate masks. The bitmask buffer is filled on CPU by xgrammar (which does need the prior token), hence the unavoidable synchronization. Then the sampler executable directly uses this buffer.

This is explained in the PR description, along with the relative performance metrics.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
carinapeng: Is this safety guarantee by way of "the constrained loop is fully sequential" (from your other comment) precisely why we are not able to do pipelining?

Since they are both about the same thing I will just consolidate the discussion here :)

carinapeng: In a structured output example the grammar contains positions to exactly one legal token like {,", :, etc. In these positions there are nothing to sample, and the token is determined by the grammar state, we don't want to be paying a full GPU inference and sampler and CPU read back of it for this

Maybe we can try forcing tokens like if exactly one bit is set, return a new `.forced(tokenId)` case; as in, we don't pay the round trip when grammar already decided the answer

stikves: _"the constrained loop is fully sequential"_

Yes, for the per-token grammar steps, inference and sampling are sequential (token `n` must be known before we can compute bitmask `n+1`). That's the inherent data dependency.

 But the "pipelined" part comes from two things:
  (1) the masking stays on GPU
  (2) jump-forward (`findJumpForwardString`) batches deterministic multi-token segments into a single prefill 

sequences like `{"name": "` encode as one GPU dispatch instead of 9 sequential round-trips

In some extreme cases (nested grammar) this matched 109% of pipelined baseline perf.

There is a solution to fully pipelining, using xgrammar's speculative decoding support (`TraverseDraftTree`). But we can only enable that when we have speculative decoding in this engine.

stikves: _"Maybe we can try forcing tokens like if exactly one bit is set, return a new .forced(tokenId) case; as in, we don't pay the round trip when grammar already decided the answer"_

For a single forced token, even if we know the answer without sampling:
(1) We still need to run inference on that token to update the KV cache and produce logits for position n+1 (we don't have a `prefill` function on the non-chunked models)
(2) We'd only save: the GPU sampler dispatch + `addCompletedHandler` callback + continuation await

But — runs of consecutive forced tokens (like `{", ": ", },\n`) are already handled by `findJumpForwardString`, which batches them into one prefill dispatch. A single isolated forced token would be much rarer.

We can leave this as a TODO, but I expect to get actual gains from speculative decode.

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
```diff
@@ -0,0 +1,232 @@
...
+        guard let constrainedEngine = inferenceEngine as? any ConstrainedGenerationCapable else {
+            throw InferenceRuntimeError.invalidArgument(
+                "PipelinedConstrainedDecodingStrategy requires a ConstrainedGenerationCapable engine")
+        }
+
+        let vocabSize = vocabSizeOverride ?? ConstrainedDecodingStrategy.deriveVocabSize(from: tokenizer)
+        guard let vocabSize else {
+            throw InferenceRuntimeError.invalidArgument(
+                "Cannot determine vocabulary size from tokenizer. "
+                    + "Pass vocabSize explicitly via CoreAIRunner or LLMAsset metadata."
```
kevchengcodes: Delete `LLMAsset`, I think this is outdated

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
```diff
@@ -0,0 +1,232 @@
...
+            )
+        }
+
+        let singleTokenStops = stopSequences.sequences.filter { $0.count == 1 }.map { $0[0] }
+        if stopSequences.sequences.contains(where: { $0.count > 1 }) {
+            CLILogger.log(
+                "Warning: Multi-token stop sequences not supported by xgrammar, using single-token stops only",
+                component: "PipelinedConstrained")
```
kevchengcodes: could be worth elevating this to always print by passing level=0

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
```diff
@@ -0,0 +1,232 @@
...
+                    with: inputTokens,
+                    samplingConfiguration: samplingConfiguration,
+                    maxTokens: maxTokens,
+                    session: session
+                )
+                self.innerIterator = stream.makeAsyncIterator()
+            }
+
+            do {
+                while true {
```
kevchengcodes, line 183: do we need any extra guards to prevent an infinite loop here?

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
```diff
@@ -0,0 +1,232 @@
...
+                    try Task.checkCancellation()
+
+                    guard var iterator = self.innerIterator else {
+                        finished = true
+                        return nil
+                    }
+
+                    guard let tokenId = try await iterator.next() else {
+                        finished = true
```
kevchengcodes, line 194: should we also set `self.innerIterator = nil`?

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
```diff
@@ -0,0 +1,232 @@
...
+                        recentTokens.removeFirst()
+                    }
+                    if stopSequences.matches(recentTokens: recentTokens) {
+                        finished = true
+                        return nil
+                    }
+
+                    // Decode text incrementally
+                    generatedTokens.append(tokenId)
+                    let fullDecode = tokenizer.decode(tokens: generatedTokens.map { Int($0) })
```
kevchengcodes, line 213: will this be re-decoding already decoded tokens per step?

swift/Sources/CoreAILanguageModels/GuidedGeneration/ConstrainedGenerationSession.swift
```diff
@@ -193,6 +193,47 @@ public struct ConstrainedGenerationSession: ~Copyable {
         matcher.reset()
         allTokensBlocked = false
     }
+
+    /// Rollback the grammar state by N tokens.
+    public mutating func rollback(_ numTokens: Int = 1) {
+        matcher.rollback(numTokens)
```
kevchengcodes: what if `numTokens` > `maxRollbackTokens`?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -1004,6 +1163,342 @@ private struct EngineImpl: ~Copyable {
...
+            bitmaskBuffer = argmax.bitmaskBuffer
+        } else {
+            throw InferenceRuntimeError.invalidArgument(
+                "Constrained generation requires MPSGraphArgmaxSampler or MPSGraphCompositeSampler")
+        }
+
+        // Pre-grow KV cache
+        let totalNeeded = prompt.count + maxTokens
+        try kvCache.ensureCapacity(forContextLength: totalNeeded, queue: pipelineQueue)
+        try logits.ensureCapacity(forContextLength: 1)
```
kevchengcodes: only 1?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -1004,6 +1163,342 @@ private struct EngineImpl: ~Copyable {
...
+                // Batch-encode jump-forward tokens + sample next token after them
+                let bitmaskPtr = bitmaskBuffer.contents().assumingMemoryBound(to: Int32.self)
+                let postJumpMask = sessionBox.fillBitmask(into: bitmaskPtr)
+                if case .terminated = postJumpMask { break }
+                let applyMaskAfterJump = postJumpMask == .constrained
+
+                lastToken = try await withCheckedThrowingContinuation { cont in
+                    do {
+                        try _encodeStepForConstrainedGeneration(
+                            tokens: jumpTokens,
```
kevchengcodes: does this need to take into account lastToken, so that this input looks like `tokens: [lastToken] + jumpTokens`

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -232,6 +236,69 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
...
+            self.acquireEngine()
+            defer {
+                self.releaseEngine()
+                self.storeConstrainedSessionForReuse(sessionBox)
+                if self._activeToken.withLock({ $0 === token }) {
+                    self._activeToken.withLock { $0 = nil }
+                    self._generationTask.withLock { $0 = nil }
+                }
+            }
+            self.engine.reset()
```
kevchengcodes, line 291: what about `history.clear()`?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -232,6 +236,69 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
...
+    /// each step, enabling constrained decoding without falling back to the sequential engine.
+    ///
+    /// The loop is semi-pipelined: inference overlaps with bitmask computation, but sampling
+    /// waits per token (grammar state is inherently sequential).
+    func generateConstrained(
+        with input: [TokenId],
+        samplingConfiguration: SamplingConfiguration,
+        maxTokens: Int,
+        session: ConstrainedSessionHandle
+    ) throws -> AsyncThrowingStream<TokenId, Error> {
```
kevchengcodes, line 263: why isn't this function async, like generate()?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
```diff
@@ -232,6 +236,69 @@ final class CoreAIPipelinedEngine: InferenceEngine, Sendable {
...
+        precondition(
```
kevchengcodes: I see the existing generate() has this. do we need it here too?
```
        if let priorTask = _generationTask.withLock({ $0 }) {
            _activeToken.withLock { $0?.cancel() }
            await priorTask.value
        }
```

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
kevchengcodes: also do we really need to reset the engine?

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
stikves: Let's do this separately

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
stikves: we depend on the inner iterator

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
stikves: decoding is unfortunately O(N^2) and cannot be avoided

swift/Sources/CoreAILanguageModels/GuidedGeneration/ConstrainedGenerationSession.swift
stikves: we now have a rollback budget that is checked

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: it was due to requirements of internal call. maybe we can do this as a follow up if needed later on

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
stikves: done

swift/Sources/CoreAILanguageModels/DecodingStrategies/PipelinedConstrainedDecodingStrategy.swift
stikves: fixed

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: fixed on prior commit

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: fixed on prior commit

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: prefix caching will be done as a follow up

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIPipelinedEngine.swift
stikves: throws instead of drain, when we have full rollback support, we can update

carinapeng:
This 2k line change is hard to review with confidence. A proposal is to land the GPU-masking capability and tests first, then the jump-forward + session cache as a follow-up

carinapeng:
> Masking happens on the GPU sampler:
https://github.com/apple/coreai-models/pull/114/changes#diff-26ae392b0ec2c8ff20901ac369417462a4f015174091b838203b599c9992db7dR239
Yes, there is a CPU work for the xgrammar, however unlike the sequential engine, it does not "stop world" to generate masks. The bitmask buffer is filled on CPU by xgrammar (which does need the prior token), hence the unavoidable synchronization. Then the sampler executable directly uses this buffer.



> > Thanks for looking into this. Getting this working is tricky because constrained decoding has some data dependency that vanilla decoding doesn't. From the grammar side, the accepted token to fill bit masking runs on CPU and the bit mask for token n+1 would depend on the sampled value of token n so this GPU-CPU roundtrip might need more design change. Happy to chat more on this and this work would probably benefit from breaking this into smaller pieces!
> 
> Masking happens on the GPU sampler: https://github.com/apple/coreai-models/pull/114/changes#diff-26ae392b0ec2c8ff20901ac369417462a4f015174091b838203b599c9992db7dR239
> 
> Yes, there is a CPU work for the xgrammar, however unlike the sequential engine, it does not "stop world" to generate masks. The bitmask buffer is filled on CPU by xgrammar (which does need the prior token), hence the unavoidable synchronization. Then the sampler executable directly uses this buffer.
> 
> This is explained in the PR description, along with the relative performance metrics.

Thanks — agreed the masking itself is on-GPU and correct, and the "no stop-the-world" framing is right: we avoid the full logit readback + CPU softmax the sequential engine pays -- this can be a PR in itself to land first

To make sure we're on the same page on the pipelining, though.. There's still a per-token sync point — the sampled token n has to be read back to CPU so acceptToken(n) can run before fillBitmask(n+1). That's the "unavoidable synchronization" you mention, and this might be why the sampled positions can't overlap. So the real speedups here are (1) GPU masking and (2) jump-forward batching the deterministic runs — not pipelining of the constrained steps

Proposal to move forward:

1. Split the PR. Land the well-contained piece first — ConstrainedGenerationCapable + GPU bitmask sampler + its unit tests. Then jump-forward + session-cache checkout/checkin as a follow-up

2. Get the real loop under test. Today the integration tests use a mock that reimplements generateConstrained, so runConstrainedCompletion + jump-forward + rollback have no automated coverage — only the manual greedy on-device check. I'd like at least one automated test that drives the actual loop through a jump-forward + rollback case before we lock the protocol shape.

3. Fix or drop stopTokenIds. It's threaded through the session inits but never reaches xgrammar (ConstrainedGenerationSession.swift:64), and the docstring claims it blocks tokens mid-generation. Either wire it in or remove it + fix the docs

stikves:
> This 2k line change is hard to review with confidence. A proposal is to land the GPU-masking capability and tests first, then the jump-forward + session cache as a follow-up

Sure, I'll send individual PRs and rebase this as they get merged. This will stay as the last piece where everything is eventually wired.

swift/Sources/CoreAILanguageModels/LanguageModel/ModelResources.swift
```diff
@@ -42,6 +42,14 @@ final class ModelResources: ResourceManaging {
         state.withLock { $0.loaded?.supportsLogits }
     }
 
+    /// Whether the loaded engine supports GPU-side constrained generation, or `nil` when unloaded.
+    var loadedEngineIsConstrainedCapable: Bool? {
```
tjia1818, line 46: what about staticShapeInferenceEngine? does this affect that?

stikves:
Closing the original as all pieces have been separated

PR #115, opened 2026-07-17
Fix Flux quality: don't quantize text encoder

### LINKED ISSUE #111: Flux quality subjectively regressed in beta v3
### What are you trying to build?

I'm working to generate playing cards using Flux and SD.

I first discussed build issues in [this issue](https://github.com/apple/coreai-models/issues/77#issuecomment-4998287310).
When I am testing the Flux model in Beta 27 compared to previous versions I feel like the quality has come down..

### Where are the current docs or utilities unclear?

Beta 3 seems to objectively lead to lower quality outputs then previous iterations.

Example prompt:

> a playful cartoon illustration of a single [Ant], [green] background, no scenery, bold clean outlines, flat shading'

Each card has the same prompt just subject and background color changed.

Image

### Expected improvement

Any ideas what to do to get better flux outputs? This seems much worse than what I would expect given how good Flux usually is.

### Additional context

_No response_

stikves:
The Flux text encoder (Qwen-based, 3B) was marked quantizable=True, causing it to be exported at INT4 when --compression 4bit is used. This degrades prompt adherence (cosine 0.9655 vs reference) because the conditioning signal loses semantic precision at 4-bit.

Set quantizable=False for the Flux text encoder so it stays at f16 regardless of global compression setting. The transformer (4B) remains INT4 — it's quality-insensitive since it just predicts noise.

Fixes #111.

### DIFF
--- a/python/src/coreai_models/diffusion/components.py
+++ b/python/src/coreai_models/diffusion/components.py
@@ -327,7 +327,7 @@ FLUX2_COMPONENTS: dict[str, ComponentSpec] = {
         output_names=("hidden_states",),
         wrapper_fn=lambda p: Flux2TextEncoderWrapper(p.text_encoder),
         dummy_fn=dummy_flux2_text_encoder,
-        quantizable=True,
+        quantizable=False,
     ),
     "vae_decoder": ComponentSpec(
         asset_name="VAEDecoder",


stikves:
We should quantize the model

PR #126, opened 2026-07-29
New custom op for KV cache update

vineet-g:
## Summary

- Unify KV-cache update + fetch into a single fused custom op
- Both iOS and macOS model definitions now use the new custom op

### DIFF
--- a/python/src/coreai_models/export/mlir_ops.py
+++ b/python/src/coreai_models/export/mlir_ops.py
@@ -13,6 +13,7 @@ Provides:
 - ``register_custom_torch_lowering`` to register all lowerings on a converter
 """
 
+import operator
 from collections.abc import Callable
 from typing import Annotated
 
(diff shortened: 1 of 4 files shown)


carinapeng:
Thank you! A high level question for the motivation of this fused op, do we have perf numbers? Or is this for correctness?

Am I right in the understanding that at the MLIR level the op is decomposed away into unsqueeze -> immutable_slice_update -> aten.slice -> aten.slice -> aten.squeeze?

Can we get the same primitives from slice_update + slice + squeeze primitives? Or is the intention to change dependency as we lower?

python/src/coreai_models/export/mlir_ops.py
```diff
@@ -149,108 +150,289 @@ def internal_func(x):  # type: ignore[no-untyped-def]
...
+            isu_node = generate_node(target_fn=immutable_slice_update)
+            isu_node = graph.node_copy(isu_node)
+            isu_node.args = (
+                _autofunc_base_tensor(autofunc_node),
+                unsqueeze_node,
+                autofunc_node.kwargs["begin"],
+                autofunc_node.kwargs["end"],
+            )
+            # 5D meta comes from the mutated-cache getitem.
+            if getitem_cache is not None:
```
carinapeng: Should we do assert rather than conditional? `getitem_cache is None` would silently drop the KV write and if we assert we can ensure cache write is mandatory

TobyRoseman:
This looks interesting. Would it be possible to add unit tests for this new functionality?

python/src/coreai_models/export/mlir_ops.py
```diff
@@ -149,108 +150,289 @@ def internal_func(x):  # type: ignore[no-untyped-def]
...
+    Appends the retired getitem nodes to ``get_items`` and records each
+    getitem-name → replacement-node mapping in ``get_item_replacements`` for the
+    caller's teardown pass.
+    """
+    for autofunc_node in cache_update_autofuncs.values():
+        # Map getitem index -> getitem node.
+        getitem_by_idx: dict[int, fx.Node] = {}
+        for user in autofunc_node.users:
+            if user.target is operator.getitem:
+                getitem_by_idx[user.args[1]] = user
```
carinapeng: So it seems that an auto functionalized custom op returns a tuple and that index 0 is the fetched slice, but to minimize brittleness and assumption that graph **always** has one get item per index, should we either assert one getitem per index, or `replace_all_uses_with` per getitem grouped by index?

vineet-g:
> Thank you! A high level question for the motivation of this fused op, do we have perf numbers? Or is this for correctness?

This is to enable KV-cache quantization. No perf change is expected; the lowered MLIR has the same primitives as before.

> Am I right in the understanding that at the MLIR level the op is decomposed away into unsqueeze -> immutable_slice_update -> aten.slice -> aten.slice -> aten.squeeze?

Yes

> Can we get the same primitives from slice_update + slice + squeeze primitives? Or is the intention to change dependency as we lower?

Existing mutable_slice_update takes one layer's K/V in and returns the whole cache out which creates asymmetry while placing quantize/dequantize nodes around the custom op. 

The new fused op takes one layer in and returns that same layer's populated cache as output which gives the quantizer a symmetric custom-op boundary.

carinapeng (approved):
Looks good, but let's make sure we are not introducing regression with the python tests, and +1 to @TobyRoseman 's point on adding unit test for this

vineet-g:
> Would it be possible to add unit tests for this new functionality?

`python/tests/test_model_units/test_primitives/test_macos/test_cache.py` already covers the unit tests for cache_update which now uses the new custom op

PR #132, opened 2026-07-31
Add state handlers for hybrid model support (>2 states)

stikves:
Both engines previously hardcoded exactly 2 states (key_cache + value_cache). Models with sliding window caches, recurrent states, or other persistent states beyond the KV pair would fail at load.

StateHandler protocol + implementations:
- GrowingNDArrayState: dynamic KV cache with capacity doubling
- FixedNDArrayState: static states (sliding caches, conv/recurrent)
- FixedMTLBufferState: MTLBuffer states for pipelined engine
- StateHandlerFactory: shape-based classification with metadata override

Engine changes:
- Sequential: uses StateHandlerFactory, supports 2-4 states
- Pipelined: additionalStates?.insertAll(into:) at all encode sites
- Both: hasNonTruncatableStates guard forces full reset for hybrid models
- LanguageConfig: optional "states" field for explicit classification

### DIFF
--- a/swift/Sources/CoreAILanguageModels/Handlers/StateHandlerFactory.swift
+++ b/swift/Sources/CoreAILanguageModels/Handlers/StateHandlerFactory.swift
@@ -33,7 +33,6 @@ struct SyncStateHandlerSet {
 /// 2. Shape-based heuristic — dynamic dim → kvCache, static + "cache" in name → slidingCache, else → fixed
 /// 3. Legacy (2 states) — both kvCache, no warning
 enum StateHandlerFactory {
-
     /// Classify states using metadata or heuristic fallback.
     static func classifyStates(
         descriptor: InferenceFunctionDescriptor,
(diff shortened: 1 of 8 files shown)


swift/Sources/CoreAILanguageModels/Handlers/StateHandlerFactory.swift
```diff
@@ -0,0 +1,178 @@
...
+                "StateHandlerFactory: \(names.count) states classified by heuristic. "
+                    + "Add \"states\" to metadata.json for explicit control.",
+                component: "StateHandlerFactory")
+        }
+
+        return classified
+    }
+
+    /// Infer state kind from shape and name.
+    private static func inferKind(name: String, descriptor: InferenceFunctionDescriptor) -> StateKind {
```
carinapeng, line 88: Are we classifying as either truncate sliding cache or fixed? Can there be a a non truncatable kv cache state?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -273,21 +256,61 @@ public final class CoreAISequentialEngine: InferenceEngine, @unchecked Sendable
...
+                states.insert(&kv0.array, for: kv0.name)
+                states.insert(&kv1.array, for: kv1.name)
+                states.insert(&s0.array, for: s0.name)
+                _ = try await function.run(
+                    inputs: [inputIdsName: inputIdsArray, positionIdsName: positionIds],
+                    states: consume states,
+                    outputViews: consume outputViews
+                )
+                handler[stateIndex: 0] = s0
+            case 2:
```
carinapeng, line 281: Is this because of the Swift ~Escapable limitation (binding N views into MutableViews before fn.run can't be a loop)? I ran into this in the other PR too

It could be worth addressing this by maybe using a buffer-based view constructor (like the pipelined `AsyncMutableValue(unsafeBuffer:)`) this way we can have generic state counts

swift/Sources/CoreAILanguageModels/Handlers/StateHandlerFactory.swift
stikves: We have 3 state types in `StateKind`

KV-Caches are always truncateable. Fixed ones can be depending on whether they are sliding or linear attention.

Standard KV-Caches are inherently truncateable by design.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
stikves: Yes, the ownership semantics does not work in a for loop. 

I think I tried that during exploration, and so far nothing except fixed lookups worked. But we can take another look in the future (assuming Swift relaxes some of these restrictions)

Lewis300 (approved):

PR #133, opened 2026-07-31
Resolve to compiled .aimodelc, when metadata references .aimodel but not present

### LINKED ISSUE #128: Compiled diffusion bundles (.aimodelc) are unloadable unless metadata.json is hand-edited; Flux2's resolveAsset masks 3 of 4 components
## Summary

Ahead-of-time compiling a diffusion model with `xcrun coreai-build compile` produces
`*.aimodelc` assets, but the `metadata.json` emitted by `coreai.diffusion.export`
still names `*.aimodel`. Nothing reconciles the two. Every diffusion load path
except three hard-coded lookups in `Flux2Pipeline` appends the metadata string to
the bundle URL verbatim, so the resulting bundle fails at first use with a
missing-file error.

For FLUX.2 this presents as "only the text encoder is broken", which is
misleading — see [Why it looks like a text-encoder bug](#why-it-looks-like-a-text-encoder-bug).

`models/README.md:173` does document the manual step ("If you compile a model,
replace the corresponding asset in the bundle directory and update `metadata.json`
to reference the new filename"), so this may be working as designed. The argument
below is that the failure mode is bad enough — silent, deferred, and partial —
to be worth fixing in code.

## What we were doing

Building a local Background Assets delivery path for FLUX.2-klein-4B on iOS, to
test on-device download against a dev build without going through App Store Connect:

1. `uv run coreai.diffusion.export flux2-klein-4b --platform iOS --overwrite`
   → `exports/FLUX.2-klein-4B/` (4.0 G; int4 per-block on transformer + text
   encoder, fp16 VAEs)
2. `xcrun coreai-build compile … --platform iOS --architecture h16g --preferred-compute gpu`
   → `compiled/FLUX.2-klein-4B/*.aimodelc` (4.0 G)
3. Stage a pack: four `.aimodelc` + `metadata.json` + `tokenizer/` + `vae_bn_*.npy`
   → `packs/flux2-klein-4b-h16g/`
4. `xcrun ba-package package manifests/flux2-klein-4b-h16g.manifest.json -o archives/flux2-klein-4b-h16g.aar`
   → 3.63 GB archive
5. `xcrun ba-serve serve … --host <LAN-IP> --port 52463`, device pointed at it
   with `xcrun ba-serve url-override`

Auditing the archive before testing the download is what surfaced this.

## The mismatch

`metadata.json` in the pack, carried over unchanged from the export:

```json
"assets": {
  "transformer_512":  "Transformer_512.aimodel",
  "text_encoder":     "TextEncoder.aimodel",
  "vae_decoder_half": "VAEDecoder_half.aimodel",
  "vae_encoder_half": "VAEEncoder_half.aimodel"
}
```

What the archive actually contains — `xcrun aa list -i … | grep -c '\.aimodel$'`
returns **0**:

```
Contents/Transformer_512.aimodelc/    Contents/TextEncoder.aimodelc/
Contents/VAEDecoder_half.aimodelc/    Contents/VAEEncoder_half.aimodelc/
Contents/metadata.json                Contents/tokenizer/
Contents/vae_bn_mean.npy              Contents/vae_bn_var.npy
```

## Mechanism

`PipelineDescriptor.loadFromMetadata` copies the `assets` values straight into
`components`, with no existence check and no extension normalization —
`PipelineDescriptor.swift:156-160`:

```swift
descriptor.components.textEncoder = assets["text_encoder"]
descriptor.components.textEncoder2 = assets["text_encoder_2"]
descriptor.components.unet = assets["transformer"] ?? assets["unet"]
descriptor.components.vaeDecoder = assets["vae_decoder"]
descriptor.components.vaeEncoder = assets["vae_encoder"]
```

Consumers append that string to the bundle URL directly:

| Call site | Components taken literally |
|---|---|
| `PipelineDescriptor+CoreAI.swift:45,48,59,109` (SD 2.x) | unet, vae_decoder, vae_encoder, text_encoder |
| `SD3Pipeline+Resources.swift:18-27` | unet, text_encoder, text_encoder_2, vae_decoder |
| `Flux2Pipeline+Resources.swift:32,81` | text_encoder, vae_encoder (`.full` mode) |
| `DiffusionRunnerMain.swift:427-433,507-516,635-638` | all |

`CoreAIDiffusionModelFunction.init` (`CoreAIDiffusionModelFunction.swift:18-20`)
only stores the URL. `AIModel(contentsOf:)` isn't called until `loadResources()`
(`:24-36`), so construction succeeds and the error surfaces later, at first use.

## Why it looks like a text-encoder bug

`Flux2Pipeline+Resources.swift:116-126` has a local helper that probes both
extensions:

```swift
private static func resolveAsset(at url: URL, name: String) -> String? {
    let fm = FileManager.default
    let aimodel = "\(name).aimodel"
    let aimodelc = "\(name).aimodelc"
    if fm.fileExists(atPath: url.appendingPathComponent(aimodel).path) { return aimodel }
    else if fm.fileExists(atPath: url.appendingPathComponent(aimodelc).path) { return aimodelc }
    return nil
}
```

It is used for the transformer (`:40,:45`), VAE decoder (`:57,:62`) and half VAE
encoder (`:83`) — all looked up by **hard-coded base name**, so `metadata.json`
never enters into it and those three resolve correctly against a compiled pack.

The text encoder has no hard-coded name to probe with; it must come from
metadata, so it goes through `descriptor.components.textEncoder` at `:32` and is
used verbatim at `:73`. One of four components fails, and only on first
text-encode.

Combined with the deferred load, a compiled FLUX.2 pack constructs cleanly,
reports mode `.half` correctly, and fails partway into the first generation. Our
previously built `packs/sd21-h16g` has the identical mismatch and would fail on
**all four** components, since SD 2.x has no `resolveAsset` equivalent.

## Reproduction

```bash
uv run coreai.diffusion.export flux2-klein-4b --platform iOS --overwrite
xcrun coreai-build compile exports/FLUX.2-klein-4B/TextEncoder.aimodel \
  --platform iOS --architecture h16g --preferred-compute gpu \
  --output compiled/TextEncoder.aimodelc
# assemble a bundle from the .aimodelc plus the untouched exported metadata.json
# → Flux2Pipeline(from:) succeeds; first text-encode fails on TextEncoder.aimodel
```

## Proposed fix

Normalize once, centrally, in `loadFromMetadata` — fixes SD 2.x, SD 3.x, FLUX.2
and `diffusion-runner` together, and makes the Flux2-local helper redundant:

```swift
/// Resolve an asset path from metadata against the bundle, tolerating a
/// `.aimodel` name that was ahead-of-time compiled to `.aimodelc`.
private static func resolveAssetPath(_ path: String?, in dir: URL) -> String? {
    guard let path else { return nil }
    let fm = FileManager.default
    if fm.fileExists(atPath: dir.appendingPathComponent(path).path) { return path }
    guard path.hasSuffix(".aimodel") else { return path }
    let compiled = path + "c"
    return fm.fileExists(atPath: dir.appendingPathComponent(compiled).path) ? compiled : path
}
```

`loadFromMetadata` already receives the `metadata.json` URL, so the bundle
directory is `url.deletingLastPathComponent()`. Falling back to the original
string when neither exists preserves today's error messages.

Alternatives, if implicit resolution is unwanted:

- **Fail eagerly.** Have `loadFromMetadata` validate declared assets and throw a
  `ModelBundle.verify()`-style `missingAsset` naming both the expected and the
  found filename. Keeps metadata authoritative; still removes the silent partial
  failure. Would need `Flux2Pipeline` to stop relying on metadata being wrong.
- **Fix it at export.** Have `coreai.diffusion.export` write a `compilation.targets`
  entry and have the compile step rewrite `assets`, so a compiled bundle is
  self-consistent on disk.

## Notes

The design question underneath this: is `metadata.json` authoritative (in which
case eager `missingAsset`-style validation in `loadFromMetadata` is the right
fix) or a hint (in which case resolution is)? The answer determines whether
`Flux2Pipeline.resolveAsset` should go away, which also touches
`bestAvailableMode`. Happy to send a patch for whichever direction you prefer —
noting the repo currently states it isn't taking code contributions
(`README.md:129`).

## Workaround

Rewrite the `assets` map when staging a compiled pack:

```bash
python3 - packs/<pack-id>/metadata.json <<'PY'
import json, sys
p = sys.argv[1]
d = json.load(open(p))
d["assets"] = {k: v.replace(".aimodel", ".aimodelc") for k, v in d["assets"].items()}
json.dump(d, open(p, "w"), indent=2)
PY
```

## Environment

- coreai-models @ `aa3bbf6`
- `coreai-torch` 0.4.1, `coreai-build` / `ba-package` / `ba-serve` from Xcode-beta
- macOS 26.6 (build 25G72), target iOS, architecture `h16g`
- Model: `black-forest-labs/FLUX.2-klein-4B`, int4 per-block

stikves:
After `coreai-build compile` produces .aimodelc from .aimodel, the metadata.json still references the original .aimodel name. This caused silent load failures for all diffusion pipelines and required manual metadata editing.

ModelBundle.resolveAssetURL now falls back from .aimodel to .aimodelc when the declared path doesn't exist on disk. 

This is one-way only (source → compiled), since the reverse would bypass an intentional user choice.

Applied to:
- ModelBundle.modelURL(for:) — LLMs, segmentation, VLMs
- ModelBundle.verify() — bundle validation
- SD3Pipeline+Resources — SD3 component loading
- PipelineDescriptor+CoreAI — SD 1.x/2.x component loading

Fixes apple/coreai-models#128

### DIFF
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/EngineFactory.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/EngineFactory.swift
@@ -25,7 +25,7 @@ public struct EngineFactory: Sendable {
     ///
     /// - Parameters:
     ///   - config: The JSON model configuration as raw data.
-    ///   - modelURL: The location of the model asset on disk.
+    ///   - modelURL: The location of the CoreAI model asset (`.aimodel` or `.aimodelc`).
     ///   - options: The engine options, including an optional variant override and KV cache
     ///     settings. Defaults to a value with auto-detection enabled and the `.auto` KV cache
     ///     strategy.
@@ -45,15 +45,12 @@ public struct EngineFactory: Sendable {
             CLILogger.log("  - kvCacheSize: \(size) (override)")
         }
 
-        // Step 2: Resolve model URL for Core AI path
-        let coreAIModelURL = PreparedModel.resolveCoreAIModelURL(from: modelURL)
-
-        // Step 3: Prepare model asset via Core AI
-        let preparedModel = try await PreparedModel.prepare(at: coreAIModelURL)
+        // Step 2: Prepare model asset via Core AI
+        let preparedModel = try await PreparedModel.prepare(at: modelURL)
 
         CLILogger.log("  - structure: \(preparedModel.structure.description)")
 
-        // Step 4: Resolve variant with structure detection
+        // Step 3: Resolve variant with structure detection
         let variant = try resolveVariant(
             override: options.variant,
             detectedStructure: preparedModel.structure
@@ -61,7 +58,7 @@ public struct EngineFactory: Sendable {
 
         CLILogger.log("  - resolved variant: \(variant.rawValue)")
 
-        // Step 5: Instantiate the appropriate engine
+        // Step 4: Instantiate the appropriate engine
         return try await selectEngine(
             variant: variant,
             config: parsedConfig,
--- a/swift/Sources/CoreAIShared/Bundle/ModelBundle.swift
+++ b/swift/Sources/CoreAIShared/Bundle/ModelBundle.swift
@@ -48,9 +48,10 @@ public struct ModelBundle: Sendable {
     }
 
     /// Resolve a component's URL within the bundle by role key.
+    /// Falls back to `.aimodelc` if the declared `.aimodel` path doesn't exist on disk.
     public func modelURL(for key: String) -> URL? {
         guard let path = assets[key] else { return nil }
-        return bundlePath.appending(path: path)
+        return Self.resolveAssetURL(path, in: bundlePath)
     }
 
     /// Required-component variant — throws `BundleError.missingField` if absent.
@@ -61,13 +62,27 @@ public struct ModelBundle: Sendable {
         return url
     }
 
+    /// Resolve an asset path against a directory, falling back from `.aimodel` to `.aimodelc`.
+    ///
+    /// When `coreai-build compile` produces a compiled `.aimodelc` from a source `.aimodel`,
+    /// metadata.json still references the original name. This finds the compiled variant
+    /// so users don't need to hand-edit metadata.json after compilation.
+    public static func resolveAssetURL(_ path: String, in directory: URL) -> URL {
+        let url = directory.appending(path: path)
+        if FileManager.default.fileExists(atPath: url.path) { return url }
+        if path.hasSuffix(".aimodel") {
+            let compiled = directory.appending(path: path + "c")
+            if FileManager.default.fileExists(atPath: compiled.path) { return compiled }
+        }
+        return url
+    }
+
     /// Verify all declared assets exist on disk. Throws `BundleError.missingAsset`
     /// with guidance if a component is missing (e.g. after manual compilation).
     public func verify() throws {
-        let fm = FileManager.default
         for (key, filename) in assets {
-            let url = bundlePath.appending(path: filename)
-            if !fm.fileExists(atPath: url.path) {
+            let url = Self.resolveAssetURL(filename, in: bundlePath)
+            if !FileManager.default.fileExists(atPath: url.path) {
                 throw BundleError.missingAsset(key: key, path: url)
             }
         }
--- a/swift/Sources/CoreAIDiffusionPipeline/Pipelines/Flux2Pipeline+Resources.swift
+++ b/swift/Sources/CoreAIDiffusionPipeline/Pipelines/Flux2Pipeline+Resources.swift
@@ -4,6 +4,7 @@
 // be found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
 
 import CoreAI
+import CoreAIShared
 import Foundation
 import Tokenizers
 
@@ -114,15 +115,9 @@ extension Flux2Pipeline {
 
     /// Resolve an asset name to a filename, checking for .aimodel or .aimodelc.
     private static func resolveAsset(at url: URL, name: String) -> String? {
-        let fm = FileManager.default
-        let aimodel = "\(name).aimodel"
-        let aimodelc = "\(name).aimodelc"
-        if fm.fileExists(atPath: url.appendingPathComponent(aimodel).path) {
-            return aimodel
-        } else if fm.fileExists(atPath: url.appendingPathComponent(aimodelc).path) {
-            return aimodelc
-        }
-        return nil
+        let resolved = ModelBundle.resolveAssetURL("\(name).aimodel", in: url)
+        guard FileManager.default.fileExists(atPath: resolved.path) else { return nil }
+        return resolved.lastPathComponent
     }
 
     /// Probe available assets and pick the highest quality mode.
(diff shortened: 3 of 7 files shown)


swift/Sources/CoreAIShared/Bundle/ModelBundle.swift
```diff
@@ -61,13 +62,27 @@ public struct ModelBundle: Sendable {
         return url
     }
 
+    /// Resolve an asset path against a directory, falling back from `.aimodel` to `.aimodelc`.
+    ///
+    /// When `coreai-build compile` produces a compiled `.aimodelc` from a source `.aimodel`,
+    /// metadata.json still references the original name. This finds the compiled variant
+    /// so users don't need to hand-edit metadata.json after compilation.
+    public static func resolveAssetURL(_ path: String, in directory: URL) -> URL {
```
tjia1818, line 70: we already have a `resolveCoreAIModelURL` in `ModelStructure.swift`, can you check?

kevchengcodes: Also a `resolveAsset` in `Flux2Pipeline+Resources.swift`, it'd be nice to unify all of these

stikves: Sure, let's try to minimize duplication.

stikves: Removed the redundant ones and forwarded to here as necessary

swift/Sources/CoreAILanguageModels/InferenceEngines/EngineFactory.swift
```diff
@@ -45,11 +45,8 @@ public struct EngineFactory: Sendable {
             CLILogger.log("  - kvCacheSize: \(size) (override)")
         }
 
-        // Step 2: Resolve model URL for Core AI path
-        let coreAIModelURL = PreparedModel.resolveCoreAIModelURL(from: modelURL)
```
kevchengcodes, line 49: we don't need a `resolveAssetURL` call here?

stikves: No,

The callers use `requireModelURL`

which calls
`return Self.resolveAssetURL(path, in: bundlePath)`

This call was redundant

swift/Sources/CoreAILanguageModels/InferenceEngines/EngineFactory.swift
```diff
@@ -45,11 +45,8 @@ public struct EngineFactory: Sendable {
...
-
-        // Step 3: Prepare model asset via Core AI
-        let preparedModel = try await PreparedModel.prepare(at: coreAIModelURL)
+        // Step 2: Prepare model asset via Core AI
+        let preparedModel = try await PreparedModel.prepare(at: modelURL)
```
tjia1818, line 49: `modelURL` here is the URL to the bundle, or `.aimodel`?

stikves: It is the `.aimodel`

tjia1818 (approved):

tjia1818 (approved):

PR #136, opened 2026-07-31
Adding Parakeet Export and Runtime

### LINKED ISSUE #7: Model request: NVIDIA Parakeet TDT 0.6B v3 (multilingual streaming ASR)
### Model

[nvidia/parakeet-tdt-0.6b-v3](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v3) (NVIDIA Parakeet TDT 0.6B v3, CC-BY-4.0)

### Why this model

The current audio catalog covers Whisper (large-v3 and large-v3-turbo), which is excellent for batch transcription but not streaming-native. Parakeet would complement it well for live use cases:

- **Streaming-friendly architecture.** TDT/RNN-T decoding is designed for incremental, low-latency output, which suits live captioning much better than sliding-window re-transcription with Whisper.
- **On-device friendly size.** At 0.6B parameters it fits comfortably in memory on iPhone-class devices, leaving headroom for the host app.
- **Multilingual.** v3 supports 25 European languages, including smaller ones (Dutch, in my case) that are poorly served by on-device alternatives.
- **Proven demand on Apple silicon.** Community CoreML ports already exist (e.g. FluidAudio), but a first-party Core AI export recipe with the ahead-of-time compilation and instant-load benefits would be a significant step up in reliability.

### Use case

I build a live-captioning app for deaf and hard-of-hearing users (real-time, fully on-device, privacy-sensitive). A streaming-capable multilingual ASR model in the Core AI catalog would directly improve accessibility apps in this category.

Thanks for considering it.

kevchengcodes:
## Summary

This PR adds the export and runtime for parakeet-tdt-0.6b-v3. It also does some light cleanup of the existing speech recognition Swift runtime, and renames all `speech-runner`-esque terminology to `speech-recognizer`. This is why you see some files deleted (they were renamed but with large changes).

Closes #7 

### Changes
- new Python file and README for export
- Swift code to differentiate between Whisper and Parakeet TDT (such as for bundle loading)
- Code addition to `MelSpectogram.swift`, such that Whisper is unaffected but Parakeet can be supported
- Parakeet TDT decoder

## Testing
[Using this jfk.wav file as an example](https://github.com/ggml-org/whisper.cpp/blob/master/samples/jfk.wav):

```bash
uv run models/parakeet/export.py --overwrite --dynamic
swift run -c release speech-recognizer exports/parakeet-tdt-0.6b-v3_float32_dynamic/ jfk.wav --verbose --warmup

### CODE REVIEW

**Review by tjia1818 — APPROVED** [review]

File: swift/Sources/CoreAIShared/Runtime/NDArray+Helpers.swift [hunk 1]
```diff
@@ -76,6 +76,37 @@ public func readNDArray<T: BitwiseCopyable>(
...
+}
+
+/// Slice overload of `fillFloatNDArray`. Lets hot-loop callers pass a slice
+/// (`buffer[a..<b]`) straight through without materializing an intermediate
+/// `Array`. This is the canonical implementation; the `[Float]` overload forwards
+/// here. Indices are taken relative to the slice's own `startIndex`.
+public func fillFloatNDArray(_ array: inout NDArray, with elements: ArraySlice<Float>) {
+    let base = elements.startIndex
+    switch array.scalarType {
+    #if !((os(macOS) || targetEnvironment(macCatalyst)) && arch(x86_64))
```
stikves, line 144: CoreAI is not supported on `x86_64`

swift/Sources/CoreAISpeech/MelSpectrogram.swift
```diff
@@ -31,43 +104,80 @@ public struct MelConfig: Sendable {
 public enum MelSpectrogram {
```
stikves, line 137: This code needs test coverage.

The audio processing is numerically sensitive.

swift/Sources/CoreAISpeech/ParakeetTDTDecoder.swift
```diff
@@ -0,0 +1,214 @@
...
+import Foundation
+
+/// Greedy decoder for Parakeet TDT.
+///
+/// Drives the autoregressive (token, duration) loop in Swift, calling the exported
+/// `decoder_step` (single LSTM step) and `joint` graphs per emission. The LSTM
+/// hidden/cell state is owned here, seeded with zeros and only advanced when the
+/// step's input token — i.e. the previous prediction — is non-blank, matching the HF
+/// `ParakeetTDTDecoderCache.update(..., mask=~blank_mask)` semantics.
+public struct ParakeetTDTDecoder: SpeechDecoder {
```
stikves, line 17: Same here, this should have extensive tests

models/parakeet/export.py
```diff
@@ -0,0 +1,418 @@
...
+
+def _convert(
+    module: torch.nn.Module,
+    example_inputs: dict[str, torch.Tensor],
+    input_names: list[str],
+    output_names: list[str],
+    dynamic_shapes: dict | None = None,
+):
+    module.eval()
+    exported = torch.export.export(
```
stikves: need to wrap in `torch.autocast`

swift/Sources/CoreAISpeech/ParakeetTDTDecoder.swift
```diff
@@ -0,0 +1,214 @@
...
+        }
+        if tEnc == 0 { return (tokens: [], stats: DecodeStats(stepTimesMs: [])) }
+
+        // For a static (padded) encoder the tail frames are zero-padding; decoding
+        // them yields spurious tokens (trailing periods). Cap the loop at the number
+        // of frames that carry real audio. Dynamic exports pass tEnc, so cap == tEnc.
+        let cap = max(1, min(tEnc, validEncoderFrames))
+
+        // Resolve descriptors for the two callable graphs.
+        guard let stepFn = try decoderStep.loadFunction(named: "main") else {
```
stikves, line 84: these should be loaded at `init`

swift/Sources/CoreAISpeech/MelSpectrogram.swift
carinapeng: +1 
Would be great to add tests for this
We could also consider having mel compute done in compiled ai model

swift/Sources/CoreAISpeech/ParakeetTDTDecoder.swift
```diff
@@ -0,0 +1,214 @@
...
+
+        var jointDecIn = NDArray(descriptor: jointDecDesc.resolvingDynamicDimensions([1, 1, hidden]))
+        var jointEncIn = NDArray(descriptor: jointEncDesc.resolvingDynamicDimensions([1, 1, hidden]))
+        var logits = NDArray(descriptor: logitsDesc.resolvingDynamicDimensions([1, 1, logitsSize]))
+
+        // Swift-side LSTM state — zero-seeded, only advanced on non-blank emissions.
+        var hState = [Float](repeating: 0, count: lstmCount)
+        var cState = [Float](repeating: 0, count: lstmCount)
+
+        var lastToken: Int32 = cfg.blankTokenId
```
carinapeng: This could be a blocker for the logic

As the model listens, most frames are "blank" (no new word), the decoder has a running memory and should only move forward when a real word comes out 

So the logic for deciding "was this a blank frame" is not enough by just checking `lastToken == blank`, this should track whether the previous step emitted a blank, not whether lastToken equals the blank id

carinapeng (commented):
Agreed with @stikves that we might want to iterate a bit more. To me it'd be great to include in this PR the decoder advancing on blank frames needs more checking, and more testing. The other things flagged are worth considering but could be follow up PRs :)

swift/Sources/CoreAIShared/Runtime/NDArray+Helpers.swift
kevchengcodes: This follows the same guard pattern that is used elsewhere in the repo as well. We can add an issue to clean these up in a follow up PR?

models/parakeet/export.py
kevchengcodes: I think because we already cast `model.to(dtype)` before doing this, that we shouldn't need the autocast. For f32 it's a no-op (and we get a warning that it isn't needed for it), and for f16 it also shouldn't be needed. But since it's functionally the same, I can add it back in if we prefer.

swift/Sources/CoreAISpeech/MelSpectrogram.swift
kevchengcodes: Added unit tests for preprocessing

swift/Sources/CoreAISpeech/ParakeetTDTDecoder.swift
kevchengcodes: Added unit tests that do not require an actual model bundle for decoder

models/parakeet/export.py
kevchengcodes: added back in

swift/Sources/CoreAISpeech/ParakeetTDTDecoder.swift
kevchengcodes: The decoder has an outer loop (steps) and an inner loop (iterations per step), and this check is inside the inner loop. I think this is the right design: within a single step, iteration x consumes the symbol iteration x-1 just emitted, so keying on "did the previous step emit a blank" would reuse a stale decoder output and skip the state advance that a real symbol requires.

However, the naming was a bit misleading so I updated some of the variable names (token -> symbol). I also tested this decoder against the Python HF implementation, which gave exact token and transcript parity.

tjia1818 (approved):

swift/Sources/CoreAISpeech/SpeechDecoder.swift
```diff
@@ -52,8 +134,11 @@ public struct WhisperDecoder: SpeechDecoder {
         var valueCache = NDArray(descriptor: valCacheNDDesc.resolvingDynamicDimensions(vcShape))
 
         var encHSArray = NDArray(descriptor: encHSNDDesc.resolvingDynamicDimensions(encoderOutputShape))
-        let encFlat = readNDArray(encoderOutput, as: Float.self, count: encoderOutputShape.reduce(1, *))
-        fillNDArray(&encHSArray, as: Float.self, with: encFlat)
+        // Both sides dispatch on the array's own scalar type: an f16 export hands back an
+        // f16 encoder output and expects an f16 `encoder_hidden_states`, so reading or
+        // filling these as `Float` would trap on the scalar-type check.
+        let encFlat = flattenAsFloat(encoderOutput)
```
stikves: Shall we use a preallocated buffer?

swift/Sources/CoreAIShared/Runtime/NDArray+Helpers.swift
```diff
@@ -121,6 +121,37 @@ public func readNDArray<T: BitwiseCopyable>(
...
+public func fillFloatNDArray(_ array: inout NDArray, with elements: [Float]) {
+    fillFloatNDArray(&array, with: elements[...])
+}
+
+/// Slice overload of `fillFloatNDArray`. Lets hot-loop callers pass a slice
+/// (`buffer[a..<b]`) straight through without materializing an intermediate
+/// `Array`. This is the canonical implementation; the `[Float]` overload forwards
+/// here. Indices are taken relative to the slice's own `startIndex`.
+public func fillFloatNDArray(_ array: inout NDArray, with elements: ArraySlice<Float>) {
+    let base = elements.startIndex
```
stikves, line 142: Also why not use `fillNDArray`?

stikves (approved):
Looks good overall, thank you

swift/Sources/CoreAIShared/Runtime/NDArray+Helpers.swift
kevchengcodes: this is just a helper function that wraps `fillNDArray`, but also does f16 vs f32 checking

swift/Sources/CoreAISpeech/SpeechDecoder.swift
kevchengcodes: This is a good find, we didn't need this intermediate array so I deleted this line, and made similar optimizations in a couple other places.

swift/Sources/CoreAISpeech/ParakeetTDTDecoder.swift
carinapeng: It seems that you did change the logic to add a tracker for previous symbol, so after a blank frame it holds the memory instead of refeeding the last word, is that correct?

kevchengcodes: yep! sorry about the confusion, I think we're saying the same thing now.

carinapeng (approved):
Looks good! :)

PR #150, opened 2026-08-05
Fix forced_decoder_ids parsing for HuggingFace generation config, and add SpeechTests target

stikves:
The Hugging Face model config will have a nested array in JSON

For example:
```json
"forced_decoder_ids": [
    [
      1,
      null
    ],
    [
      2,
      50360
    ]
  ],
```
(from https://huggingface.co/openai/whisper-large-v3/blob/main/generation_config.json)


Package.swift
```diff
@@ -266,6 +266,11 @@ let package = Package(
                 .linkedLibrary("c++")
             ]
         ),
+        .testTarget(
```
tjia1818, line 269: let's make the PR title include this fix

tjia1818: or is this intended to be included in this PR at all?

stikves: We did not have tests for Speech, this adds a new target.

Yes, we can include this in the title

tjia1818 (approved):

PR #155, opened 2026-08-05
Check for end-of-turn tokens in tokenizer config top-level

Lewis300:
- Check for end-of-turn tokens in tokenizer config top-level, some HF models ship with them there.
- Add "eot_token"


stikves (approved):
Thanks! And maybe add a test to match?

PR #156, opened 2026-08-06
Add bind(into:) for zero-copy state binding

stikves:
State handlers are now classes that own their NDArrays at refcount 1.

`bind(into:)` inserts all states into `MutableViews` in a loop using stdlib `_overrideLifetime` to express disjoint
element access. Same pattern for `AsyncMutableViews` on the pipelined engine's `FixedMTLBufferState`.

## Changes

- `SyncStateHandler` protocol: add `bind(into:)`, constrain to `AnyObject`
- `FixedNDArrayState`, `GrowingNDArrayState`: struct → class, dictionary-backed
- `FixedMTLBufferState`: struct → class, add `bind(into:)` for `AsyncMutableViews`
- `runWithStates()`: sequential engine helper
- `encodeWithStates()`: pipelined engine helper
- Sequential engine: replace inline state binding with `runWithStates()`
- Pipelined engine: replace 3 inline state binding blocks with `encodeWithStates()`

Requires `Lifetimes` experimental Swift feature.
Tested with 2 state and 4 state hybrid models.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/Handlers/StateHandler.swift
+++ b/swift/Sources/CoreAILanguageModels/Handlers/StateHandler.swift
@@ -8,16 +8,13 @@ import CoreAI
 /// Persistent model state that the engine carries across inference steps.
 ///
 /// Each handler manages one or more named state tensors (e.g., key_cache + value_cache,
-/// or a single convolution_state). The engine owns an array of handlers and delegates
-/// all state lifecycle to them — allocation, growth, and reset.
+/// or a single convolution_state). The engine owns handlers and delegates all state
+/// lifecycle to them — allocation, growth, and reset.
 ///
-/// ## MutableViews Lifetime Constraint
-///
-/// `InferenceFunction.MutableViews.insert` creates a lifetime dependency on each `inout`
-/// variable. This means state binding CANNOT be abstracted into a method that takes
-/// `inout MutableViews` — the inserts must happen in the same scope as `function.run()`.
-/// Handlers therefore expose their arrays directly via subscript for the engine to bind.
-public protocol SyncStateHandler {
+/// Handlers are classes (AnyObject) so they own their NDArrays at refcount 1 —
+/// `bind(into:)` calls `mutableRawView()` without triggering COW. The loop uses
+/// `_overrideLifetime` to express disjoint element access to the compiler.
+public protocol SyncStateHandler: AnyObject {
     /// Names of the states managed by this handler.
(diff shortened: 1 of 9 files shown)


swift/Sources/CoreAILanguageModels/Handlers/StateHandler.swift
```diff
@@ -35,18 +32,32 @@ public protocol SyncStateHandler {
...
+    func truncate(to tokenCount: Int)
+}
+
+// MARK: - Lifetime Helpers
+
+/// Detach lifetime dependencies from MutableViews so it can cross scope
+/// boundaries (closures, await). Caller must ensure inserted arrays remain valid.
+@inline(__always)
+@_unsafeNonescapableResult
+@_lifetime(immortal)
```
carinapeng, line 58: "immortal" 😄

swift/Sources/CoreAILanguageModels/Handlers/StateHandler.swift
```diff
@@ -35,18 +32,32 @@ public protocol SyncStateHandler {
...
+func _unsafeEscapeMutableViews(
```
carinapeng, line 59: "immortal" sounds extreme to me... is this the best approach? We wanna tell the compiler that these views live forever?

swift/Tests/LanguageModelsTests/StateHandlerTests.swift
```diff
@@ -88,3 +88,75 @@ struct StateHandlerConformanceTests {
...
+        }
+    }
+
+    func reset() {}
+    func truncate(to tokenCount: Int) {}
+}
+
+@Suite("bind(into:) Tests")
+struct BindTests {
+    @Test("binds 1 through 4 states into MutableViews")
```
carinapeng, line 130: Do we still have a runtime limit to 4 states max?

swift/Sources/CoreAILanguageModels/Handlers/StateHandler.swift
stikves: Yes, unfortunately there is no in between now. We are opening a careful escape hatch that allows us to manually control lifetime.

swift/Tests/LanguageModelsTests/StateHandlerTests.swift
stikves: We have only tested up to 4 states.

tjia1818 (approved):

PR #160, opened 2026-08-10
State handling for static-shape inference engine

carinapeng:
## Purpose
Static-shape LLM inference engine (StaticShapeEngine) built on the `bind(into:)` state binding from #156. States are discovered by name and the KV cache is right-sized per context bucket; ctx is parsed from the function name. Input preparation is pluggable per model family via StaticInputProvider + StaticModelProfile; StaticInputContext composes the shared InputContext and the provider mirrors SyncInputHandler.

Testing: LanguageModelsTests (317) pass; correct output on static assets.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
+++ b/swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
@@ -41,9 +41,34 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
     // Largest query length across all extend functions — used as prefill threshold.
     private let maxQueryLength: Int
 
-    // Fixed size caches shared across all decoding functions.
-    private var keyCache: NDArray
-    private var valueCache: NDArray
+    // Generic persistent state (KV + any hybrid/sliding states), discovered by name from the
+    // descriptor and zero-initialized. No hardcoded KV contract, no per-model state properties.
+    //
+    // Each static graph is compiled with its own per-ctx sequence strides, so a state buffer
+    // MUST be physically laid out for the exact bucket the running graph expects — a max-ctx
+    // allocation viewed through a slice hands the smaller-bucket graph the wrong stride and
+    // corrupts the cache. We therefore RIGHT-SIZE the ctx-scaled states (e.g. the global KV
+    // cache) to the running bucket and grow+re-lay-out on a bucket crossing (`ensureStateCtx`).
+    // States whose descriptor is constant across buckets (e.g. a fixed sliding-window ring)
+    // are allocated once and never re-laid-out.
+    private var stateHandler: FixedNDArrayState
+    private let stateNames: [String]
+    // ctx bucket -> that bucket's (name, descriptor) list, for re-allocation on crossing.
+    private let stateDescriptorsByCtx: [Int: [(name: String, descriptor: NDArrayDescriptor)]]
+    // States whose sequence extent scales with the ctx bucket (must be re-laid-out on grow).
+    private let ctxScaledStateNames: Set<String>
+    // The ctx bucket the state buffers are currently laid out for.
+    private var currentStateCtx: Int
+    // Sorted ascending ctx buckets available in the model.
+    private let stateCtxLadder: [Int]
+
+    // Model-specific input preparation (embedding gather, position ids, causal mask, step, and
+    // profile extras like PLE / dual-RoPE / sliding-ring). The engine never learns an input name.
+    private let providers: [any SyncInputHandler]
+    private let profileName: String
+
+    // Bundle location (for the model profile to find sidecar resources, e.g. the PLE table).
+    private let bundleURL: URL
 
     // Number of tokens already processed in the current sequence.
     public private(set) var processedTokenCount: Int = 0
@@ -64,10 +89,11 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
 
     // MARK: - Initialization
 
-    public init(configuration: ModelConfig, preparedModel: PreparedModel) async throws {
+    public init(configuration: ModelConfig, preparedModel: PreparedModel, bundleURL: URL) async throws {
         self.config = configuration
         self.model = preparedModel.model
         self.functions = [:]
+        self.bundleURL = bundleURL
 
         let allNames = model.functionNames
         CLILogger.log("Model loaded: \(allNames.count) functions: \(allNames.sorted())")
@@ -94,7 +120,10 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
         var largestContextExtend: (name: String, descriptor: InferenceFunctionDescriptor)?
         for name in extendFunctionNames {
             let desc = try Self.requireDescriptor(model: model, functionName: name)
-            if Self.contextLength(descriptor: desc, config: configuration) == configuration.maxContextLength {
+            let ctx =
+                Self.parseContextLength(functionName: name)
+                ?? Self.contextLength(descriptor: desc, config: configuration)
+            if ctx == configuration.maxContextLength {
                 largestContextExtend = (name, desc)
                 break
             }
@@ -104,32 +133,174 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
                 "Failed to find an extend function with the max context length of \(configuration.maxContextLength)")
         }
 
-        // Validate output/state contract against the max-context function
+        // Validate output contract against the max-context function
         try Self.validateIOContract(descriptor: largestExtendDescriptor, functionName: largestExtendName)
 
         // Load embeddings
         self.embeddingTable = try await Self.loadEmbeddingTable(from: model)
 
-        // Allocate KV cache IOSurfaces sized to the max-context descriptor
-        if case .ndArray(let keyCacheDescriptor) = largestExtendDescriptor.stateDescriptor(of: Self.keyCacheName),
-            case .ndArray(let valueCacheDescriptor) = largestExtendDescriptor.stateDescriptor(of: Self.valueCacheName)
-        {
-            self.keyCache = NDArray(descriptor: keyCacheDescriptor)
-            self.valueCache = NDArray(descriptor: valueCacheDescriptor)
-            CLILogger.log(
-                "KV cache allocated: key \(keyCacheDescriptor.minimumByteCount) bytes, value \(valueCacheDescriptor.minimumByteCount) bytes (IOSurface)"
-            )
-        } else {
-            throw InferenceRuntimeError.invalidState(
-                "No KV cache state descriptors found — cannot allocate cache buffers")
+        // Build the per-ctx-bucket state descriptor table. Each extend bucket declares the
+        // same state NAMES but with its own physical shape/strides; we allocate against the
+        // exact bucket the running graph uses (see the field docs above).
+        var descriptorsByCtx: [Int: [(name: String, descriptor: NDArrayDescriptor)]] = [:]
+        for name in extendFunctionNames where name.hasPrefix("extend") {
+            let desc = try Self.requireDescriptor(model: model, functionName: name)
+            let ctx =
+                Self.parseContextLength(functionName: name)
+                ?? Self.contextLength(descriptor: desc, config: configuration)
+            guard descriptorsByCtx[ctx] == nil else { continue }
+            var descs: [(name: String, descriptor: NDArrayDescriptor)] = []
+            for stateName in desc.stateNames {
+                guard case .ndArray(let d) = desc.stateDescriptor(of: stateName) else {
+                    throw InferenceRuntimeError.invalidState(
+                        "State '\(stateName)' has no NDArray descriptor")
+                }
+                descs.append((stateName, d))
+            }
+            descriptorsByCtx[ctx] = descs
         }
+        guard !descriptorsByCtx.isEmpty else {
+            throw InferenceRuntimeError.invalidState("No extend function to size the state caches from")
+        }
+        let ctxLadder = descriptorsByCtx.keys.sorted()
+        self.stateCtxLadder = ctxLadder
+        self.stateDescriptorsByCtx = descriptorsByCtx
 
-        CLILogger.log("Engine initialized")
+        // Classify each state as ctx-scaled (its sequence extent changes across buckets → must
+        // be re-laid-out on a bucket crossing) or fixed (constant across buckets → allocated
+        // once). Derived by comparing the smallest and largest bucket's descriptor for each state.
+        let smallestCtx = ctxLadder.first!
+        let largestCtx = ctxLadder.last!
+        var ctxScaled: Set<String> = []
+        let smallDescs = descriptorsByCtx[smallestCtx]!
+        let largeDescs = Dictionary(
+            uniqueKeysWithValues: descriptorsByCtx[largestCtx]!.map { ($0.name, $0.descriptor) })
+        for (name, smallDesc) in smallDescs {
+            if let largeDesc = largeDescs[name], largeDesc.shape != smallDesc.shape {
+                ctxScaled.insert(name)
+            }
+        }
+        self.ctxScaledStateNames = ctxScaled
+
+        // Allocate at the smallest bucket; ctx-scaled states grow on demand via ensureStateCtx.
+        self.stateHandler = FixedNDArrayState(states: smallDescs)
+        self.stateNames = smallDescs.map(\.name)
+        self.currentStateCtx = smallestCtx
+        CLILogger.log(
+            "States allocated at ctx=\(smallestCtx): \(stateNames) "
+                + "(ctx-scaled: \(ctxScaled.sorted()))")
+
+        // Assemble input providers (base + model profile) and verify — at load — that every
+        // declared graph input is produced by some provider. Fail closed on any gap.
+        let (providers, profileName) = try StaticModelProfile.assembleInputHandlers(
+            descriptor: largestExtendDescriptor, bundleURL: bundleURL)
+        self.providers = providers
+        self.profileName = profileName
+
+        try Self.checkInputCoverage(
+            providers: providers, model: model, extendFunctionNames: extendFunctionNames)
+
+        CLILogger.log("Engine initialized (profile: \(profileName), \(providers.count) input providers)")
+    }
+
+    /// Ensure the ctx-scaled state buffers are physically laid out for context bucket `ctx`,
+    /// growing + re-laying-out the written prefix when a decode step crosses into a larger
+    /// bucket. Fixed states (constant descriptor across buckets, e.g. a sliding-window ring)
+    /// are carried over verbatim. No-op when already laid out for `ctx`.
+    private func ensureStateCtx(_ ctx: Int) throws {
+        guard ctx != currentStateCtx, ctxScaledStateNames.isEmpty == false else { return }
+        guard let targetDescs = stateDescriptorsByCtx[ctx] else {
+            throw InferenceRuntimeError.invalidState(
+                "No state descriptors for ctx bucket \(ctx) (ladder: \(stateCtxLadder))")
+        }
+        let old = stateHandler
+        var new = FixedNDArrayState(states: targetDescs)
+        let copyPositions = min(processedTokenCount, currentStateCtx)
+        for i in stateNames.indices {
+            let name = stateNames[i]
+            var dst = new[stateIndex: i]
+            let src = old[stateIndex: i]
+            if ctxScaledStateNames.contains(name) {
+                // Grow: copy the written sequence prefix into the larger layout.
+                Self.copyStatePrefix(from: src.array, to: &dst.array, copyPositions: copyPositions)
+            } else {
+                // Fixed: identical shape across buckets — carry the whole buffer over.
+                Self.copyStatePrefix(from: src.array, to: &dst.array, copyPositions: nil)
+            }
+            new[stateIndex: i] = dst
+        }
+        stateHandler = new
+        currentStateCtx = ctx
+        CLILogger.log("State caches re-laid-out: ctx \(currentStateCtx) → \(ctx) (copied \(copyPositions) positions)")
+    }
+
+    /// Copy a state buffer's written sequence prefix from `src` into the (larger) `dst` layout.
+    /// The sequence dim is the last dim; `copyPositions == nil` copies the whole buffer (fixed
+    /// states, identical shape). Honors the cache's channel-interleave: physically the layout is
+    /// `[groups, ctx, factor]`, so positions `[0, copyLen)` across the `factor` interleaved
+    /// channels form one contiguous `copyLen·factor` run per group. src and dst share group order
+    /// + interleave, differing only in ctx — so a per-group run copy is correct.
+    private static func copyStatePrefix(from src: NDArray, to dst: inout NDArray, copyPositions: Int?) {
+        let srcShape = src.shape
+        guard srcShape.isEmpty == false else { return }
+        let dstShape = dst.shape
+        let seqDim = srcShape.count - 1
+        let srcSeq = srcShape[seqDim]
+        let dstSeq = dstShape[seqDim]
+        let factor = src.interleaveLayout?.factor ?? 1
+        let copySeq = min(copyPositions ?? srcSeq, min(srcSeq, dstSeq))
+        guard copySeq > 0 else { return }
+        let groupCount = srcShape.reduce(1, *) / srcSeq / factor
+        let srcGroupStride = srcSeq * factor
+        let dstGroupStride = dstSeq * factor
+        let runElems = copySeq * factor
+
+        func run<T: BitwiseCopyable>(_ type: T.Type) {
+            let srcView = src.view(as: T.self)
+            var dstView = dst.mutableView(as: T.self)
+            dstView.withUnsafeMutablePointer { dstPtr, _, _ in
+                srcView.withUnsafePointer { srcPtr, _, _ in
+                    for g in 0..<groupCount {
+                        dstPtr.advanced(by: g &* dstGroupStride).update(
+                            from: srcPtr.advanced(by: g &* srcGroupStride), count: runElems)
+                    }
+                }
+            }
+        }
+        switch src.scalarType {
+        case .float16, .bfloat16: run(Float16.self)
+        case .float32: run(Float.self)
+        case .int8: run(Int8.self)
+        default: run(Float16.self)
+        }
+    }
+
+    /// Every input any extend/prompt graph declares must be produced by exactly one handler,
+    /// except `transformer_input` / `embedding_table`, which the engine produces directly.
+    private static func checkInputCoverage(
+        providers: [any SyncInputHandler], model: AIModel, extendFunctionNames: [String]
+    ) throws {
+        var declared = Set<String>()
+        for name in extendFunctionNames {
+            if let desc = model.functionDescriptor(for: name) {
+                declared.formUnion(desc.inputNames)
+            }
+        }
+        // Engine-produced inputs (embedding gather needs the model's gather graph).
+        let engineProduced = declared.filter { $0.contains("transformer_input") || $0 == "embedding_table" }
+        var produced = providers.reduce(into: Set<String>()) { $0.formUnion($1.inputNames) }
+        produced.formUnion(engineProduced)
+        let uncovered = declared.subtracting(produced)
+        guard uncovered.isEmpty else {
+            throw InferenceRuntimeError.invalidState(
+                "No input handler produces required input(s): \(uncovered.sorted()). "
+                    + "Add them to the model profile (StaticModelProfile.resolve).")
+        }
     }
 
     public convenience init(configuration: ModelConfig, modelURL: URL) async throws {
         let preparedModel = try await PreparedModel.prepare(at: modelURL)
-        try await self.init(configuration: configuration, preparedModel: preparedModel)
+        try await self.init(configuration: configuration, preparedModel: preparedModel, bundleURL: modelURL)
     }
 
     // MARK: - Initialization Helpers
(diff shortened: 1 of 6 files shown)


swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -94,7 +120,10 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
         var largestContextExtend: (name: String, descriptor: InferenceFunctionDescriptor)?
         for name in extendFunctionNames {
             let desc = try Self.requireDescriptor(model: model, functionName: name)
-            if Self.contextLength(descriptor: desc, config: configuration) == configuration.maxContextLength {
+            let ctx =
+                Self.parseContextLength(functionName: name)
```
kevchengcodes, line 119: is there an advantage to introducing this method to get context length from the function name? Or was the original `Self.contextLength(descriptor: desc, config: configuration)` not sufficient anymore?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +133,174 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+        // Build the per-ctx-bucket state descriptor table. Each extend bucket declares the
+        // same state NAMES but with its own physical shape/strides; we allocate against the
+        // exact bucket the running graph uses (see the field docs above).
+        var descriptorsByCtx: [Int: [(name: String, descriptor: NDArrayDescriptor)]] = [:]
+        for name in extendFunctionNames where name.hasPrefix("extend") {
+            let desc = try Self.requireDescriptor(model: model, functionName: name)
+            let ctx =
+                Self.parseContextLength(functionName: name)
+                ?? Self.contextLength(descriptor: desc, config: configuration)
+            guard descriptorsByCtx[ctx] == nil else { continue }
```
kevchengcodes, line 146: can we instead make sure `extendFunctionNames` is a set at this point, so there's no need for this check? Maybe we can just make the input a set instead of a list.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +133,174 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+                }
+                descs.append((stateName, d))
+            }
+            descriptorsByCtx[ctx] = descs
+        }
+        guard !descriptorsByCtx.isEmpty else {
+            throw InferenceRuntimeError.invalidState("No extend function to size the state caches from")
+        }
+        let ctxLadder = descriptorsByCtx.keys.sorted()
+        self.stateCtxLadder = ctxLadder
```
kevchengcodes: Why store this? I don't really see it used

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +133,174 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+        self.stateDescriptorsByCtx = descriptorsByCtx
+
+        // Classify each state as ctx-scaled (its sequence extent changes across buckets → must
+        // be re-laid-out on a bucket crossing) or fixed (constant across buckets → allocated
+        // once). Derived by comparing the smallest and largest bucket's descriptor for each state.
+        let smallestCtx = ctxLadder.first!
+        let largestCtx = ctxLadder.last!
```
kevchengcodes: could also just do descriptorsByCtx.keys.min() and .max()

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +133,174 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+                            from: srcPtr.advanced(by: g &* srcGroupStride), count: runElems)
+                    }
+                }
+            }
+        }
+        switch src.scalarType {
+        case .float16, .bfloat16: run(Float16.self)
+        case .float32: run(Float.self)
+        case .int8: run(Int8.self)
+        default: run(Float16.self)
```
kevchengcodes: is this ok default behavior if a different scalar type is passed?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -380,6 +517,10 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
         var logitBuffer = [LogitsScalarType](repeating: 0, count: config.vocabSize)
         var currentPosition = processedTokenCount
 
+        if processedTokenCount == 0 {
+            stateHandler.reset()
```
kevchengcodes, line 522: can you add comment why it is reset here?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -393,41 +534,60 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
 
             CLILogger.log("Graph: \(graphName), batch=\(batchSize), step=\(batchStartToken), tokens=\(tokensInBatch)")
 
+            let fn = try loadFunction(named: graphName)
+            let desc = try functionDescriptor(for: graphName)
+            let graphInputs = Set(desc.inputNames)
```
kevchengcodes, line 540: this looks to be in the while loop. Do we need to reload the function each iteration?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -393,41 +534,60 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
-                tokensInBatch: tokensInBatch
-            )
+                batchSize: batchSize,
+                slidingWindow: nil,
+                descriptors: descriptors)
+
+            var inputs: [String: NDArray] = [:]
+            // Embedding gather is engine-produced (needs the model's gather graph).
+            if let txName = desc.inputNames.first(where: { $0.contains("transformer_input") }) {
+                if graphInputs.contains("embedding_table") { inputs["embedding_table"] = embeddingTable }
```
kevchengcodes: I'm seeing transformer_input and embedding_table hardcoded in a couple places now. Can you confirm this is safe and scalable?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -393,41 +534,60 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+            // lifetime to the handler; a stored-property borrow would end at the call).
+            let handler = stateHandler
             var states = InferenceFunction.MutableViews()
-            states.insert(keyCacheView, for: Self.keyCacheName)
-            states.insert(valueCacheView, for: Self.valueCacheName)
+            handler.bind(into: &states)
             var outputs = try await fn.run(
                 inputs: inputs,
-                states: consume states,
+                states: _unsafeEscapeMutableViews(consume states),
```
kevchengcodes, line 591: is this unavoidable?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -544,7 +637,7 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
         }
 
         var tokenArray = NDArray(descriptor: tokenNDDesc)
-        let tokenView = tokenArray.mutableView(as: Int32.self)
+        var tokenView = tokenArray.mutableView(as: Int32.self)
```
kevchengcodes, line 641: is `tokenView` modified downstream?

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,395 @@
...
+        let profile = try resolve(descriptor: desc, bundleURL: bundleURL)
+        providers.append(contentsOf: profile.extraHandlers)
+        return (providers, profile.name)
+    }
+
+    static func resolve(descriptor desc: InferenceFunctionDescriptor, bundleURL: URL) throws -> StaticModelProfile {
+        let inputs = Set(desc.inputNames)
+        let states = Set(desc.stateNames)
+
+        // Gemma4: PLE + dual-RoPE + sliding-window KV cache.
```
kevchengcodes: Are Gemma4 and Qwen3.5 the only models supported in this workflow?

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,395 @@
...
+        // block rather than the first that merely exists.
+        var lang: [String: Any]? = nil
+        for url in candidates where FileManager.default.fileExists(atPath: url.path) {
+            guard let data = try? Data(contentsOf: url),
+                let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
+                let l = json["language"] as? [String: Any]
+            else { continue }
+            lang = l
+            break
+        }
```
kevchengcodes: can this be simplified? Every LLM has the same ModelBundle structure so we can just look for the correct metadata.json instead of parsing through candidates?

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/// `rope_cos` / `rope_sin` — Gemma4 dual RoPE, precomputed on host and passed as inputs (lets
+/// the model exceed the ANE 16-bit position-id limit). The per-dim theta vector is computed once;
+/// per step we fill `cos(pos·θ)` / `sin(pos·θ)` for each block position.
+///
+/// Sliding sub-range `[0, slidingHeadDim)` = GPT-NeoX full rotary (freqs repeated in two halves).
+/// Global sub-range = partial rotary: only the first `floor(partialRotaryFactor·globalHeadDim/2)`
+/// frequency pairs rotate; the rest are NoPE (θ = 0 ⇒ cos = 1, sin = 0).
+final class RoPEProvider: SyncInputHandler {
```
kevchengcodes: could use some better test coverage of the classes defined in this file

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: The descriptor version uses `keyCache.shape.max()`, which is wrong when a fixed dim exceeds a small bucket

A example is qwen3's cache being [..,1024,..256], so at the 256 bucket .max() returns 1024 and collapses buckets. `extend_<ctx>_<q>` can be the source of truth here, so I parse it and keep descriptor as a fallback, this is covered by a test!

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Good catch, this was only used in one error string, removed it and the error sorts inline

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: True! Done

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Changed it to preconditionFailure so an unexpected scalar type fails loudly instead of being silently reinterpreted as fp16 👍

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Yes done! See line 528–531

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: loadFunction is a cached lazy load (`functions[name]` dict) — first call loads, the rest are dict lookups

Not keeping it out of for loop because graphName changes across buckets, but there is no repeated compile

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: They are graph I/O contract like logits/key_cache, but I lifted them to be constants (transformerInputToken, embeddingTableName) along side others. 

The engine produces them directly because embedding gather needs the model's gather graph so they can't just be SyncInputHandlers

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Yes `MutableViews` is `~Escapable` and `bind(into:)` ties its lifetime to the handler so passing it to fn.run across that boundary needs the escape hatch 

It's the sam pattern the shared runWithSates uses in https://github.com/apple/coreai-models/pull/156

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: It has to be var because the point here is we are writing the token ids into a buffer and writing into buffer counts as mutating the view even if we never reassign tokenView variable itself

It is the same pattern as other handlers

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: Those are what's tested, but the profile isn't hardcoded to them — resolve() is a pure function of the graph's declared inputs/states (base + RoPE / sliding-window / PLE), so a new model whose graph declares those inputs works with no engine change.

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: Good point, I replaced it with ModelBundle(at:), changed the directory scanning logic

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: Added unit tests for the context-bucket parsing and causal-mask fill 👍 

I looked at the per provider output tests (RoPE/PLE, etc) too but they need a graph NDArrayDescriptor to size their output, and that type is only produced by a real export so these are tested by the end to end exports

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: The dedup we need is by context, not by name. A set de-dups by name but the names are already unique here, guard is doing the work that a Set can't do here

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
kevchengcodes: Is it ok to use it as a fallback then, if the fallback can potentially give a wrong answer?

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticInputProviders.swift
```diff
@@ -0,0 +1,100 @@
...
+// input descriptors) so it can size its own buffer to the active bucket. Embedding gather
+// is produced by the engine directly (it needs the model's gather graph), so there is no
+// handler for `transformer_input`.
+
+/// `position_ids` — absolute UInt16 positions across the aligned block.
+final class PositionIdsProvider: SyncInputHandler {
+    private let name: String
+    var inputNames: [String] { [name] }
+    init(name: String) { self.name = name }
+    func prepare(_ ctx: InputContext) async throws -> [String: NDArray] {
```
blevine1, line 23: I have concerns about the performance of this whole `prepare`/`InputHandler` design. Storing the handlers as existentials and making new NDArrays per inference is expensive. And also the design of having each handler for every IO be a separate object, stored in an array of existentials, each allocating a new dictionary per query... compounds the issue. These heavy levels of abstraction have notable costs and I'm curious what motivated such a complex design? 

Having a simple function that takes in the already allocated IO buffers and populates them would be simpler and far more performant.

carinapeng: This is a fair concern, the existential array and the `prepare -> [String: NDArray]` is from the `InputHandler` design which this PR adopts, the allocation is from this shared abstraction. I think it is outside of scope of this PR to change that at this point

Two things I can do without changing the contract, we can cache the per-bucket input descriptors and reuse the per bucket input buffers across decode steps

The deeper design change an be a follow up. I'm happy to put together the buffer populating approach if it's helpful to compare

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -41,9 +46,32 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
     // Largest query length across all extend functions — used as prefill threshold.
     private let maxQueryLength: Int
 
-    // Fixed size caches shared across all decoding functions.
-    private var keyCache: NDArray
-    private var valueCache: NDArray
+    // Generic persistent state (KV + any hybrid/sliding states), discovered by name from the
+    // descriptor and zero-initialized. No hardcoded KV contract, no per-model state properties.
+    //
+    // Each static graph is compiled with its own per-ctx sequence strides, so a state buffer
```
blevine1: Giving a bit of a perspective from someone coming up to speed on this code: This comment has a lot of interesting details, but is also really hard to parse on a first pass when trying to become familiar with the engine. This seems to refer to really specific implementation details, and terminology like "RIGHT-SIZE", or "bucket", which at least I don't know what they are referring to in this context (and suspect others wouldn't either).

Consider making up-front comments on fields like this more concise. Like:
```
The state values used by this engine. 
The states may be reallocated over the engine's lifetime depending on what strategy the engine/model uses for sizing  the values as context grows.
```
And then leave it to individual usage sites to have clear code that make it easy to understand what's happening. This will make the code more approachable IMO and also avoid the trap that I think this comment is falling into of over-describing how the field is currently used by code that is nowhere in sight, which will also age poorly as implementations change and be easy to forget to update.

If there is actually a desire for an enumeration of all the different "KV cache growing strategies" used by different engines in a single location, that might be better left for a separate design doc that can also have pictures. Not for an inline comment.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -41,9 +46,32 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+    private let stateNames: [String]
+    // ctx bucket -> that bucket's (name, descriptor) list, for re-allocation on crossing.
+    private let stateDescriptorsByCtx: [Int: [(name: String, descriptor: NDArrayDescriptor)]]
+    // States whose sequence extent scales with the ctx bucket (must be re-laid-out on grow).
+    private let ctxScaledStateNames: Set<String>
+    // The ctx bucket the state buffers are currently laid out for.
+    private var currentStateCtx: Int
+
+    // Model-specific input preparation (embedding gather, position ids, causal mask, step, and
+    // profile extras like PLE / dual-RoPE / sliding-ring). The engine never learns an input name.
```
blevine1:
```suggestion
    // The handlers which produce/populate the model inputs.
```

The comment as written enumerates too many specific details that will age poorly. Also the phrase "The engine never learns an input name" reads like a weird platitude

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +136,177 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+            descriptorsByCtx[ctx] = descs
+        }
+        guard !descriptorsByCtx.isEmpty else {
+            throw InferenceRuntimeError.invalidState("No extend function to size the state caches from")
+        }
+        self.stateDescriptorsByCtx = descriptorsByCtx
+
+        // Classify each state as ctx-scaled (its sequence extent changes across buckets → must
+        // be re-laid-out on a bucket crossing) or fixed (constant across buckets → allocated
+        // once). Derived by comparing the smallest and largest bucket's descriptor for each state.
```
blevine1, line 164: Curious for my own knowledge, what are the actual type of states that all of this abstraction is being used to generalize for? 

Please lmk if any docs I should read

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +136,177 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+        self.providers = providers
+        self.profileName = profileName
+
+        try Self.checkInputCoverage(
+            providers: providers, model: model, extendFunctionNames: extendFunctionNames)
+
+        CLILogger.log("Engine initialized (profile: \(profileName), \(providers.count) input providers)")
+    }
+
+    /// Ensure the ctx-scaled state buffers are physically laid out for context bucket `ctx`,
```
blevine1: If the function was named like `growStatesIfNeeded(for: context)` then this long comment could be avoided I think. Best to prefer self-documenting code when possible.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +136,177 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+        for i in stateNames.indices {
+            let name = stateNames[i]
+            var dst = new[stateIndex: i]
+            let src = old[stateIndex: i]
+            if ctxScaledStateNames.contains(name) {
+                // Grow: copy the written sequence prefix into the larger layout.
+                Self.copyStatePrefix(from: src.array, to: &dst.array, copyPositions: copyPositions)
+            } else {
+                // Fixed: identical shape across buckets — carry the whole buffer over.
+                Self.copyStatePrefix(from: src.array, to: &dst.array, copyPositions: nil)
```
blevine1: Confused here, why are we doing any work if we don't need to grow the state buffer? It's not usable as is?

It looks like this code is making the new `FixedNDArrayState` thing even if no states actually grow?

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -104,32 +136,177 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+        let dstSeq = dstShape[seqDim]
+        let factor = src.interleaveLayout?.factor ?? 1
+        let copySeq = min(copyPositions ?? srcSeq, min(srcSeq, dstSeq))
+        guard copySeq > 0 else { return }
+        let groupCount = srcShape.reduce(1, *) / srcSeq / factor
+        let srcGroupStride = srcSeq * factor
+        let dstGroupStride = dstSeq * factor
+        let runElems = copySeq * factor
+
+        func run<T: BitwiseCopyable>(_ type: T.Type) {
```
blevine1, line 250: It's unfortunate you have to implement this yourself, we should expose an API for copying from one NDArray view to another. We have such capability but didn't get an API out for it. Consider filing a feedback 😄

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,378 @@
...
+// Use of this source code is governed by a BSD-3-clause license that can
+// be found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+import CoreAI
+import CoreAIShared
+import Foundation
+
+// MARK: - Static Model Profile
+
+/// A model family's contribution to the static-shape engine: the extra input providers it
```
blevine1: "A model family's contribution to the static shape engine"

This phrase reads strange to me and I don't follow what it means.

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,378 @@
...
+/// needs beyond the universal decoder set. Selected as *data* from the descriptor's declared
+/// inputs/states — no per-model engine subclass, no `matches(model:)` returning an engine.
+struct StaticModelProfile {
```
blevine1, line 16: What does the term 'profile' mean in this name?

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,378 @@
...
+                    let v = isSin ? Foundation.sin(angle) : Foundation.cos(angle)
+                    ptr[query &* strides[1] &+ d &* strides[2]] = LogitsScalarType(v)
+                }
+            }
+        }
+    }
+}
+
+/// `ple_embeddings` — per-layer embedding INT8 rows, gathered per token from an mmap'd sidecar.
+final class PLEProvider: SyncInputHandler {
```
blevine1: What's the motivation for this being a class instead of a struct?

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
```diff
@@ -0,0 +1,378 @@
...
+        guard let d = ctx.descriptors["ple_embeddings"] else { return [:] }
+        var ple = NDArray(descriptor: d)
+        try table.gather(tokenIDs: Array(ctx.tokens), batchSize: ctx.batchSize, into: &ple)
+        return ["ple_embeddings": ple]
+    }
+}
+
+/// mmap'd per-layer embedding table (`*_ple.safetensors`, INT8 `[vocab, rowWidth]`), with a
+/// row-major gather. Bounds-checked so no token id can index past the mapping.
+final class PerLayerEmbeddings {
```
blevine1: What's the motivation for this being a class instead of a struct?

blevine1 (commented):
The changes looks good to me in general given the pre-existing design, but I'd consider zooming out and potentially changing some details of the "input handler" structure/interface at a later point. It is a pretty thick abstraction that'll limit what performance can be achieved.

Left some comments mostly about structure and readability.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Agreed, the comment was giving unnecessary details, trimmed comment and moved details to usage site!

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Addressed

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: Renamed to `growStatesIfNeeded(for: context)` and dropped the long comment

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: The states are the KV cache, plus for hybrid models a fixed depth sliding window KV ring, DeltaNet conv/recurrent states. Only the global KV grows with context, thus `ctx-scaled`, the sliding ring and cone/recurrent states are fixed-size. We don't have public doc but if this abstraction grows we can consider adding one

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: This is a good catch, it runs only on an actual context growth (guarded by `ctx != currentStateCtx` and non-empty ctx-scaled set) so something always grows when it's entered.

But you're right the fixed states were being copied for nothing — changed it to reuse the fixed states' existing buffers as-is (no realloc/copy)

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
carinapeng: :) Yeah — added a TODO referencing it and we can file the feedback for an NDArray view-to-view copy API

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: I agree. Dropped "profile" as jargon, and reworded to "The set of input handlers a model needs, chosen from its declared graph inputs/states."

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: Same as above on the rewording

carinapeng: The name just refers to the per-model handler set (base + RoPE/PLE/sliding)

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: No particular reason — they're stateless value holders. Changed them (and the other handlers) to struct

swift/Sources/CoreAILanguageModels/InferenceEngines/StaticModelProfile.swift
carinapeng: Addressed as above

carinapeng:
@blevine1 Thanks Ben! Agree with your comments, the SyncInputHandler abstraction (existentials + prepare -> [String: NDArray]) is inherited, so I'd like to keep the interface change out of this PR. I'll file a follow up to revisit the handler structure / interface. Appreciate the review!

carinapeng:
Still iterating on design, will close this for now

PR #147, opened 2026-08-05
Add SyncInputHandler protocol for engine extensibility

stikves:
## Summary

Composable input preparation layer for inference engines, enabling model-specific inputs (RoPE, PLE, sliding window masks) without engine modifications.

- **SyncInputHandler** protocol with `prepare(_ context:) -> [String: NDArray]`
- **InputContext** with `.dynamic()` and `.static()` builders
- **TokenInputHandler** — standard token/position ID preparation with batch caching
- **CompositeInputHandler** — wraps base handler with extra model-specific inputs
- **PipelinedTokenInputHandler** — MTLBuffer-based for pipelined engine
- **InputCoverage.verify()** — fail-fast at engine init if inputs are uncovered

Also hardens `NDArray+Helpers`:
- `fillNDArray(count:using:)` and `readNDArray` now check contiguity and fall back to stride-aware indexing for tensors with GPU alignment padding

## Test plan

- [x] InputHandler Tests: 4 tests pass (context builders, type conformance, coverage API)
- [x] CoreAIShared Tests: 24 tests pass (no regressions)
- [x] Full package builds cleanly

PR #162, opened 2026-08-10
Extension to InputContext design: adding per-graph descriptors

carinapeng:
## Summary

Adds a `descriptors: [String: NDArrayDescriptor]` field to `InputContext` (from #147) so an input handler can size its output buffer to the running graph's shape. Backward-compatible: defaults to empty, and the dynamic path is unchanged.

## Why

Static-shape / bucketed engines select a different graph per step, so a handler (position ids, causal mask, RoPE, PLE, ...) needs the active graph's input descriptors to allocate correctly. `InputContext` didn't carry them.

## Change

- New `descriptors` field; threaded through `.dynamic()` (empty) and `.static()` (optional, defaults empty).
- No change to the `SyncInputHandler` protocol or existing callers.

Enabling change for the static-shape engine (PR to follow).


carinapeng:
Still iterating on design, will close this for now

PR #166, opened 2026-08-11
Refactor macOS export logic | Retrieve reference inputs / names / dynamic shapes from model class.

Lewis300:
- Moves the macOS export **graph contract** — which inputs exist, which are state, what they're called, how they're shaped — out of `export/macos.py` and onto the model class as overridable hooks.
- **No behavior change** for anything that exports today: all 8 registered macOS models produce bit-identical reference inputs, shape bounds, and graph names.
- Fixes two real bugs found on the way: exports with a context ≤ 2048 crashed inside `torch.export`, and activation calibration read the cache length off a guessed input position.

## The hooks

```python
export_input_names()  -> {graph: (name, ...)}
export_state_names()  -> {graph: (name, ...)}
export_output_names() -> {graph: (name, ...)}
build_reference_inputs(config, target_dtype, spec) -> {graph: {param: tensor}}
build_dynamic_shapes(config, spec)                 -> {graph: shapes}
```

Keyed by graph name — a macOS model has one graph, `main`. The defaults are what `macos.py` hardcoded, so `export_macos_model` and the pipeline's quantization step now just call them, and `validate_export_contract` cross-checks that all five agree.

Two ordering rules: **reference inputs** bind to the traced signature, so they must be in its exact order; **name lists** are looked up by name, so each carries only the relative order of its own kind.

iOS classes inherit the macOS-shaped defaults, which are wrong in every respect, so they raise until the iOS follow-up implements them.

### DIFF
--- a/python/src/coreai_models/export/compression.py
+++ b/python/src/coreai_models/export/compression.py
@@ -166,13 +166,16 @@ def quantize_pytorch_model(
 
         reset_positions = set(state_indices)
         for pos in reset_positions:
-            assert pos >= 2, (
-                "States cannot occupy the first two input positions. "
-                "Those must be reserved for input_ids and position_ids"
-            )
-            assert pos < len(inputs), (
-                f"State index out of bounds, got {pos}, while the number of inputs is {len(inputs)}"
-            )
+            if pos < 2:
+                raise ValueError(
+                    "States cannot occupy the first two input positions. "
+                    "Those must be reserved for input_ids and position_ids"
+                )
+            if pos >= len(inputs):
+                raise IndexError(
+                    f"State index out of bounds, got {pos}, while the number of inputs is "
+                    f"{len(inputs)}"
+                )
 
         # Match the caller's declared bound: position_ids.shape[1] <= cache_seq_len - 1
         # (the `seq_pos` Dim in `BaseForCausalLM.build_dynamic_shapes`).
--- a/python/src/coreai_models/export/compression.py
+++ b/python/src/coreai_models/export/compression.py
@@ -11,15 +11,18 @@ using the coreai-opt library, including calibration data preparation.
 """
 
 import logging
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 
 import torch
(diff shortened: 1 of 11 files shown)


tjia1818 (approved):

gokulkrishna98 (approved):

python/src/coreai_models/models/base.py
```diff
@@ -7,28 +7,84 @@
...
+    def __post_init__(self) -> None:
+        # Below this there is no legal `Dim(min=query_len, max=max_context_length - 1)`.
+        if self.max_context_length < self.query_len + 2:
+            raise ValueError(
+                f"max_context_length={self.max_context_length} is too small to trace: "
+                f"it must be at least query_len + 2 = {self.query_len + 2}."
+            )
+        assert self.cache_seq_len <= self.max_context_length, (
+            "cache_seq_len must not be greater than max_context_length. Received "
+            f"cache_seq_len = {self.cache_seq_len}, "
```
carinapeng: This is inconsistent with the raise ValueError four lines up for the too-small-context case, suggest raise ValueError instead

python/src/coreai_models/export/compression.py
```diff
@@ -151,9 +164,18 @@ def quantize_pytorch_model(
...
         device = next(model.parameters()).device
 
-        cache_seq_len = inputs[2].shape[-2]
-        # Match the dynamic-shape upper bound declared by the caller:
-        #   position_ids.shape[1] <= cache_seq_len - 1   (see pipeline.py `seq_pos` Dim)
+        reset_positions = set(state_indices)
+        for pos in reset_positions:
+            assert pos >= 2, (
+                "States cannot occupy the first two input positions. "
+                "Those must be reserved for input_ids and position_ids"
```
carinapeng: Same as above, ie `state_indices=(0,)` would zero `input_ids` during calibration instead of erroring. Recommend `raise ValueError/IndexError`

carinapeng (approved):
Comments about error handling, everything else looks good. Thanks!

PR #171, opened 2026-08-14
Fix VLM export for single-images

stikves:
StaticVisionEncoder rejected `num_frames=1` when `temporal_patch_size=2` because `1 % 2 != 0`.

But the single-image path in `_patchify` already handles this by duplicating across the temporal dimension, so `grid_t=1` is correct.

Only enforce the divisibility check for multi-frame exports.


carinapeng (approved):

python/src/coreai_models/vlm/export.py
```diff
@@ -638,7 +641,7 @@ async def export_vision_encoder(
     ).eval()
     del hf_model
 
-    grid_t = num_frames // spec.temporal_patch_size
+    grid_t = 1 if num_frames == 1 else num_frames // spec.temporal_patch_size
```
kevchengcodes: this is calculated during `StaticVisionEncoder.init`. Can we just reuse it from there?
```python
grid_t = wrapper.grid_t
```

PR #175, opened 2026-08-15
Add SmolLM2 support (1.7B, 360M, 135M)

stikves:
Add HuggingFace SmolLM2 model family for macOS and iOS export:
- 1.7B: INT4 quantized with FP16 embedding (tied weights)
- 360M/135M: FP16 (small enough uncompressed)

SmolLM2 uses the Qwen2 architecture under a different model_type. Adds _model_type_override field to ModelPreset to route export without registering a global type mapping.

Custom quantization config keeps embedding/lm_head at FP16 because SmolLM2 ties these weights — INT4 on the shared tensor degrades long-form generation quality (+26% PPL vs +15% with FP16 embedding).


models/smollm2/smollm2_1_7b_6bit.yaml
```diff
@@ -0,0 +1,14 @@
+# 6-bit palettization for SmolLM2-1.7B iOS.
+# 64 centroids — middle ground between 4-bit (16) and 8-bit (256).
```
pkmandke, line 2: we can skip the second line comment

```suggestion
# 6-bit palettization for SmolLM2-1.7B iOS.
```

stikves:
Will reopen after quality improvements

PR #179, opened 2026-08-16
Extract lastSafeIndex into shared free function

stikves:
Both ThinkTagParser and ToolCallParser had identical private implementations of the streaming marker hold-back logic. 

Pull itinto a module-internal free function in StreamingMarkerMatcher.swift to eliminate the duplication.

### DIFF
--- a/swift/Sources/CoreAILanguageModels/StreamingMarkerMatcher.swift
+++ b/swift/Sources/CoreAILanguageModels/StreamingMarkerMatcher.swift
@@ -1,0 +1,24 @@
+// Copyright 2026 Apple Inc.
+//
+// Use of this source code is governed by a BSD-3-clause license that can
+// be found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+/// Returns the rightmost index in `buffer` such that the suffix from that
+/// index to the end is NOT a non-empty prefix of `tag`.
+///
+/// Used by streaming parsers to decide how much of a buffer can be safely
+/// emitted without cutting off a marker that might span two deltas. At most
+/// `tag.count - 1` trailing characters are held back (the longest partial
+/// prefix that could still complete on the next delta).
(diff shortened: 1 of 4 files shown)


swift/Sources/CoreAILanguageModels/StreamingMarkerMatcher.swift
kevchengcodes, line 1: We only have 1 source file at the top level of `CoreAILanguageModels`: `ToolCallParser.swift`. Would this fit cleanly into one of the existing sub-folders, or could we add it to `ToolCallParser.swift`? Just for cleaner code organization

tjia1818 (approved):

swift/Tests/LanguageModelsTests/StreamingMarkerMatcherTests.swift
```diff
@@ -0,0 +1,37 @@
+// Copyright 2026 Apple Inc.
+//
+// Use of this source code is governed by a BSD-3-clause license that can
+// be found in the LICENSE file or at https://opensource.org/licenses/BSD-3-Clause
+
+import Testing
+
+@testable import CoreAILanguageModels
+
+@Suite("StreamingMarkerMatcher")
```
kevchengcodes, line 10: some extra tests worth considering: buffer = '', '<', '>', '!', and other combinations of escape and special characters.

swift/Sources/CoreAILanguageModels/StreamingMarkerMatcher.swift
stikves: This should be fine for now. We can iterate as more code comes here

swift/Tests/LanguageModelsTests/StreamingMarkerMatcherTests.swift
stikves: Yes. I'll try to add some more tests

PR #184, opened 2026-08-19
Live Transcription with Parakeet v3

### LINKED ISSUE #164: SpeechRecognitionModel estimates validEncoderFrames proportionally when the subsampling arithmetic determines it exactly
### Model name

parakeet-tdt-0.6b-v3

### Command run

```shell
None. Found by code inspection of swift/Sources/CoreAISpeech/SpeechRecognitionModel.swift at commit 82a0e5e5. If the field refuses to be empty, paste exactly that sentence.
```

### macOS / iOS target

macOS 27.0 beta (build 26A5406e)

### Xcode version

Xcode 27.0 beta (build 27A5194q)

### Python / uv version

Python 3.13.14

### Full error output

```shell
None. Found by code inspection of swift/Sources/CoreAISpeech/SpeechRecognitionModel.swift at commit 82a0e5e5.
```

### Anything else?

**What happens.** In `swift/Sources/CoreAISpeech/SpeechRecognitionModel.swift` (lines 254 to 269 at commit 82a0e5e5), `validEncoderFrames` is computed as a proportional estimate, `round(validMelFrames / totalMelFrames * tEnc)` clamped to `[1, tEnc]`, and the doc comment notes a possible one-frame error. That estimated value then caps the TDT decode loop in `ParakeetTDTDecoder.swift` (line 139), and `SpeechParityTest.swift` (lines 132 to 146) has to tolerate a one-frame drift against the reference before failing, which slightly weakens the parity harness.

**Why it can be exact.** The FastConformer front end reduces time resolution through three stride-2 subsampling stages, each computing `T = (T - 1) / 2 + 1` in integer arithmetic. Applying that formula three times to `validMelFrames` gives the exact number of encoder frames that correspond to real audio, with no proportionality assumption. The subsampling factor is already surfaced in the export metadata (`encoder.subsampling_factor`), so the runtime has what it needs to compute this deterministically, or the export could carry the per-stage formula explicitly if you prefer not to hard-code the stage count.

**Impact.** The decode cap can land one frame early or late around real audio boundaries, which occasionally clips or admits a trailing token, and the parity test's one-frame tolerance exists only to absorb this estimate. Replacing the estimate with the exact arithmetic tightens both, and would let the parity harness compare the full valid region without slack.

kevchengcodes:
## Summary

This PR introduces a streaming mode for Parakeet to enable live transcription via buffered/chunked inference. Offline transcription Is largely unaffected and retains the same behavior/accuracy as before, with some additive, non-breaking public API changes.

The primary use case for Parakeet streaming is through the Swift APIs for app integration. Optionally, the CLI exposes the streaming interface for diagnostic and correctness testing purposes.

**NOTE:** Streaming is for streaming bundles ONLY. Static and dynamic bundles will be rejected.

  ### Changes
- new export flags to size an export suitable for chunked encoder and streaming decoder inference. Streaming bundles will include a streaming section in their metadata
- `SpeechRecognitionModel` extension with streaming APIs
- new `StreamingSession.swift` and `StreamingWindow.swift` files, modeled off of NeMo's ASR examples
- `updates to ParakeetTDTDecoder` with new `Stream` class, built on top of existing decoder loop
- new flags in CLI `speech-recognizer` tool to invoke streaming with static audio files (intended for diagnostic use only)
- some NDArray helper functions for small allocation optimizations
- Parakeet struggles to transcribe speech, followed by a long pause, followed by more speech. The speech after the pause will often get missed for a few seconds. To workaround this, this PR adds a reset to the predictor after a long pause: `EndpointingConfig.resetAfterSilenceFrames`. This is a divergence from NeMo.
- small fixes that closes #164

## Testing

See `/models/parakeet/README.md` for comments on how to integrate this into a sample application. For CLI testing:

```bash
uv run models/parakeet/export.py --streaming --dtype float16
swift run -c release speech-recognizer --model exports/parakeet-tdt-0.6b-v3_float16_streaming150 --audio-path <audio.wav> --verbose --stream
```

Users can optionally modify the parameters in `EndpointingConfig`. The rest of the geometry gets read directly from the bundle's metadata. 

The CLI includes a `deferred-decode` flag (NeMo's "simulated" flag) for correctness testing. See the tool's `--help` for more details.

There are also new unit tests under `SpeechTests` target.


models/efficient-sam/README.md
```diff
@@ -59,7 +59,7 @@ uv run export.py --help
 ### In your iOS and macOS applications
 
 ```swift
-import ImageSegmenter
+import CoreAIImageSegmenter
```
tjia1818: good catch, can we separate this (and SAM3 readme) into a different PR?

swift/Sources/CoreAISpeech/NDArrayScan.swift
```diff
@@ -0,0 +1,116 @@
+// Copyright 2026 Apple Inc.
```
tjia1818: why we do not put these into NDArray+Helpers.swift

tjia1818 (approved):

swift/Sources/CoreAISpeech/NDArrayScan.swift
kevchengcodes: moved

PR #187, opened 2026-08-20
Add OpenAI-compatible LLM server

stikves:
Adds `llm-server` — a standalone OpenAI-compatible HTTP server for CoreAI models.
Standard eval harnesses (lm-eval, etc.) can connect to compiled models without custom integration.

```bash
llm-server --model exports/qwen3_4b_4bit_dynamic --port 8080

# Then from any OpenAI-compatible client:
OPENAI_API_KEY=local lm_eval --model openai-chat-completions \
  --model_args model=local,base_url=http://localhost:8080/v1/chat/completions \
  --tasks gsm8k --limit 10
```

### Endpoints

| Endpoint | Method | Notes |
|----------|--------|-------|
| `/v1/chat/completions` | POST | Streaming + non-streaming |
| `/v1/completions` | POST | Per-token logprobs (sequential engine) |
| `/v1/models` | GET | Model list |
| `/health` | GET | Liveness probe |

### Architecture

- **CoreAILMCommon** library: shared OpenAI-compatible API types
- **llm-server** executable: Hummingbird v2 HTTP server + handlers

### Additional changes

- Batched `forcedContinuation`: process all tokens in one forward pass for faster perplexity evaluation
- `ThinkTagParser.stripCompleted`: single-pass O(n) strip for `--no-thinking` mode
- `--raw-tokens`: accept pre-tokenized input

### Design

- Hummingbird v2 (Apache-2.0) for HTTP
- Single-request concurrency (returns 429 for overlapping calls)
- Logprobs and guided generation require `--variant coreai-sequential`
- `--no-thinking` strips reasoning blocks from output

### Tests

- 30 API types contract tests (LMCommonTests)
- ThinkTagParser tests
- Validated with lm-eval (GSM8K, IFEval, HellaSwag, MMLU Pro)


stikves:
Sample run transcript

```bash
 % ./internal/test_server.sh 
Server: /Users/sukru/Library/Developer/Xcode/DerivedData/coreai-models-mine-cfiaeetxesopcycaaniszryvqzpp/Build/Products/Release/llm-server
Model:  exports/qwen3_0_6b_4bit_dynamic
Port:   8090

Starting server...

⏳ Preparing AI asset from source... done in 0.000s (cache hit)

Serving qwen3_0_6b_4bit_dynamic on http://127.0.0.1:8090

2026-08-21T16:06:07-0700 info Hummingbird: [HummingbirdCore] Server started and listening on 127.0.0.1:8090
Server ready (PID 68023)

=== /health ===
  PASS: health endpoint
=== /v1/models ===
  PASS: models list
=== Chat: What is 2+2? ===
[coreai-1] 29t → 6t in 0.23s (25.7 tok/s)
  PASS: chat completion
  Response: 

2.
=== Chat streaming: Count to 3 ===
[coreai-2] stream done: 18 tokens (stop)
  PASS: streaming response
  PASS: streaming DONE
=== Chat with stop ===
[coreai-3] 19t → 10t in 0.19s (52.9 tok/s)
  PASS: stop sequence respected
=== Error: empty messages ===
[coreai-4] 11t → 10t in 0.06s (173.5 tok/s)
  PASS: handles empty messages

=== Results: 7 passed, 0 failed ===
All good.

Stopping server...

```

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -530,6 +554,63 @@ extension CoreAISequentialEngine.GenerationSequence {
...
+
+                // Split into per-position logit vectors.
+                // Skip the prompt positions (inputTokens.count - 1 positions);
+                // we want logits that predict each forced token.
+                let promptLen = inputTokens.count
+                var buffer: [[LogitsScalarType]] = []
+                for i in 0..<forced.count {
+                    let offset = (promptLen - 1 + i) * vocabSize
+                    let endOffset = offset + vocabSize
+                    guard endOffset <= allLogits.count else { break }
```
carinapeng: If this break ever fires we end up with `buffer.count < forced.count`, but generation still runs for `maxTokens` steps, so the fast path will index past the end of buffer later on. A short buffer means the logic and token counts disagree, so I'd throw rather than breaking silently

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -530,6 +554,63 @@ extension CoreAISequentialEngine.GenerationSequence {
...
+                step += 1
+                if step >= maxTokens {
+                    stopReasonStore.setIfUnset(.maxTokens)
+                    finishAndRelease()
+                }
+                return InferenceOutput(tokenId: token, logits: logits)
+            }
+
+            // First call with forcedContinuation + logits: batch-process all tokens at once.
+            if let forced = forcedContinuation, returnsLogits, step == 0 {
```
carinapeng, line 571: This batched path is trickiest bookkeeping, can we please add a test on it?

swift/Sources/Tools/llm-server/CompletionHandler.swift
```diff
@@ -0,0 +1,207 @@
...
+        return .init(index: index, text: "", logprobs: nil, finishReason: "stop")
+    }
+
+    guard allTokens.count <= state.config.maxContextLength else {
+        throw ServerError.badRequest(
+            "Prompt length \(allTokens.count) exceeds max context \(state.config.maxContextLength)")
+    }
+
+    let paddingToken = allTokens[0]
+    let continuation = Array(allTokens.dropFirst()) + [paddingToken]
```
carinapeng, line 139: Can you please add a comment explaining the padding token for consumers of the code (and agents)? We want to make sure lm-eval never sums the trailing entry when aligning by the offset

swift/Sources/Tools/llm-server/CompletionHandler.swift
```diff
@@ -0,0 +1,207 @@
...
+            promptText: promptText, index: idx,
+            wantsEcho: wantsEcho, topN: topN, state: state
+        )
+        choices.append(choice)
+    }
+
+    let elapsed = ContinuousClock.now - t0
+    let seconds = Double(elapsed.components.seconds) + Double(elapsed.components.attoseconds) / 1e18
+    let totalTokens = req.prompt.reduce(0) { acc, p in
+        if p.hasPrefix("__TOKEN_IDS__:") {
```
carinapeng: Let's use enum? So we have one helper and format is defined in one single place

swift/Sources/Tools/llm-server/ChatHandler.swift
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// MARK: - Helpers
+
+private func tokenizeMessages(_ messages: [ChatMessage], state: ServerState) -> [Int] {
+    var templateMessages: [[String: any Sendable]] = []
+    for msg in messages {
+        var content = msg.content.textContent
+        if msg.role == "system" && state.config.noThinking {
+            content += "\n/no_think"
```
carinapeng, line 371: Is this Qwen specific? Do other model families inject smth into the system prompt? It'd be great to gate on model family to ensure correctness

swift/Sources/Tools/llm-server/ChatHandler.swift
```diff
@@ -0,0 +1,375 @@
...
+            var lastCleanedLength = 0
+            var tokenCount = 0
+
+            for try await result in tokenStream {
+                fullText += result.text
+                tokenCount += 1
+                let cleaned = stripThinkingTags(fullText)
+
+                if cleaned.count > lastCleanedLength {
+                    let delta = String(cleaned.dropFirst(lastCleanedLength))
```
carinapeng: This assumes that models always show thinking up front, is this always a valid assumption?

carinapeng (commented):
A note on the API shape, since this is a public-ish API, is that it is easy to get things wrong w/ `__TOKEN_IDS__` when we pack magic prefixed strings inside a prompt field across diff files, we could use a `Prompt` enum (.text / .tokenIds)

Another question on the scope, do we envision this to be for a local eval server?  Or do we see this being shared by diff independent clients concurrently? Could this infra support that or is that intended?

swift/Sources/Tools/llm-server/ChatHandler.swift
stikves: `/no_think` as a fallback would work on most models. We can do a helper with more patterns if needed later on.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
stikves: Makes sense, let's throw

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
stikves: Adding a test

swift/Sources/Tools/llm-server/CompletionHandler.swift
stikves: Enum makes sense

swift/Sources/Tools/llm-server/ChatHandler.swift
stikves: All supported models emit thinking tokens up front.

tjia1818 (approved):

PR #188, opened 2026-08-20
Add StaticInputHandler protocol for zero-allocation input filling

stikves:
Adds `InputBuffers` and `StaticInputHandler`, a protocol for filling pre-allocated input buffers in-place without per-step allocation.

### Design

```swift
public protocol StaticInputHandler: Sendable {
    var inputNames: [String] { get }
    func registerBuffers(into buffers: inout InputBuffers)
    func fill(_ context: InputContext, into buffers: inout InputBuffers) throws
}
```

- Engine owns `InputBuffers`, creates once at init, passes `inout` each step
- `_modify` accessor on `InputBuffers` subscript avoids COW (26× vs get/set)
- Non-mutating `fill` — handler holds no per-step state
- Backward compatible: existing `SyncInputHandler` unchanged

Primary benefit is architectural (no COW risk, compile-time ownership via inout) rather than raw throughput.


tjia1818 (approved):

swift/Sources/CoreAILanguageModels/Handlers/InputHandler.swift
```diff
@@ -88,3 +94,88 @@ public enum InputCoverage {
...
+public struct InputBuffers {
+    private var buffers: [String: NDArray] = [:]
+    private var descriptors: [String: NDArrayDescriptor] = [:]
+
+    public init() {}
+
+    /// Register a named input with its descriptor. Allocates initial buffer at batch=1.
+    public mutating func register(name: String, descriptor: NDArrayDescriptor) {
+        descriptors[name] = descriptor
+        let resolved = descriptor.resolvingDynamicDimensions([1, 1])
```
carinapeng: Hard blocker is here

We get a crash because register is always substituting [1, 1] but static extend inputs have fixed dimensions - position_id is [1, 16] so this would give the error which we saw "Shape at dimension 1 of 1 is not a valid substitution for source shape 16" for Qwen3 0.6b

The fix would be allocating straight from the resolved descriptor, or only substitute dimensions that are actually dynamic (-1) and leave fixed dimensions alone

Also note register hardcodes a 2-D [1, 1] regardless of rank, the causal mask is 4D and step is 1D so it's wrong for those too

swift/Sources/CoreAILanguageModels/Handlers/InputHandler.swift
```diff
@@ -88,3 +94,88 @@ public enum InputCoverage {
...
+        buffers[name] = NDArray(descriptor: resolved)
+    }
+
+    /// Ensure the named buffer has capacity for the given shape.
+    /// Only reallocates if the current shape doesn't match.
+    public mutating func ensureCapacity(name: String, shape: [Int]) {
+        guard let desc = descriptors[name] else { return }
+        let current = buffers[name]
+        let resolved = desc.resolvingDynamicDimensions(shape)
```
carinapeng: Same fixed-dim issue as register — resolvingDynamicDimensions can't substitute into a fixed static dimension

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -42,28 +42,19 @@ public final class CoreAISequentialEngine: InferenceEngine, @unchecked Sendable
     private let functionDescriptor: InferenceFunctionDescriptor
 
     // I/O names from descriptor
-    private let inputIdsName: String
-    private let positionIdsName: String
     private let logitsName: String
 
+    // Input handling — handler owns allocation and fill logic
+    private var inputHandler: TokenInputHandler
```
carinapeng, line 52: Another blocker

The sequential engine uses TokenInputHandler, so the new TokenStaticInputHandler which you are adding is unused.

Either wire the sequential path to TokenStaticInputHandler so we get in place fill, or drop it

swift/Sources/CoreAILanguageModels/Handlers/InputHandler+Token.swift
```diff
@@ -97,3 +97,53 @@ public struct CompositeInputHandler<Base: SyncInputHandler>: SyncInputHandler {
...
+
+    public func registerBuffers(into buffers: inout InputBuffers) {
+        buffers.register(name: inputIdsName, descriptor: inputIdsDescriptor)
+        buffers.register(name: positionIdsName, descriptor: positionIdsDescriptor)
+    }
+
+    public func fill(_ context: InputContext, into buffers: inout InputBuffers) throws {
+        let tokens = context.tokens
+        let batchSize = tokens.count
+        precondition(batchSize > 0, "TokenStaticInputHandler: empty token batch")
```
carinapeng: prefer throwing InferenceRuntimeError over a precondition that crashes on an empty batch

swift/Tests/LanguageModelsTests/InputFillerTests.swift
carinapeng, line 1: The unit tests build InputBuffers from hand-made NDArray(shape:) buffers, so they never exercise register/ensureCapacity with a real static descriptor — which is exactly the path that crashes

A test that registers from an actual descriptor would have caught the crash before device testing, please add one

carinapeng (changes requested):
There are two blockers, suggest testing more and testing end to end before marking as review ready

For perf/CoW, I agree with you that maybe we want to switch to _modify

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
stikves: We now moved to `StaticBucketInputFiller`, the other design had one unavoidable copy

swift/Tests/LanguageModelsTests/InputFillerTests.swift
stikves: Unfortunately `NDArrayDescriptor` initializer is private, and there is no way to properly test this.

Any "hack" we do would be just deceptive as we would only test helpers, not actual descriptors.

swift/Sources/CoreAILanguageModels/Handlers/InputHandler.swift
stikves: Yes, we shouldn't have dynamic shapes in static engine. This got mixed in from the sequential one.

swift/Sources/CoreAILanguageModels/Handlers/InputHandler+Token.swift
stikves: This entire code is now deleted

swift/Sources/CoreAILanguageModels/Handlers/InputHandler.swift
stikves: removed like the other

stikves:
> There are two blockers, suggest testing more and testing end to end before marking as review ready
> 
> For perf/CoW, I agree with you that maybe we want to switch to _modify

They should be handled now, please take a look

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -124,6 +128,50 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+            $0.contains("step") && !$0.contains("pos")
+        })
+
+        // Build per-bucket descriptors from ALL extend/prompt functions
+        var bucketDescs: [StaticBucketInputFiller.BucketKey: StaticBucketInputFiller.BucketDescriptors] = [:]
+        for name in extendFunctionNames {
+            let desc = try Self.requireDescriptor(model: model, functionName: name)
+            let ql = Self.queryLength(descriptor: desc, functionName: name)
+            // Parse context length from function name (extend_256_64 → 256)
+            let nameParts = Array(name.split(separator: "_").suffix(2))
```
carinapeng: This context-length parsing is reimplemented 4 times with diff fallbacks. Let's extract one helper function and use it everywhere this way we can have one helper implemented and tested and not having different divergence at runtime

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -124,6 +128,50 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
 
+        // Discover input names from the largest-context function
+        let posName = largestExtendDescriptor.inputNames.first(where: { $0.contains("pos") })
+        guard let positionIdsName = posName else {
+            throw InferenceRuntimeError.invalidState("No position_ids input found in model descriptor")
+        }
+        let maskName: String? =
+            largestExtendDescriptor.inputNames.contains("causal_mask")
+            ? "causal_mask" : nil
+        let stepInputName: String? = largestExtendDescriptor.inputNames.first(where: {
```
carinapeng: This is fragile in how we match input names with `contains()`. Preferred to use exact names like how we do for `causal_mask` or even better a centralized name resolution step

swift/Sources/CoreAILanguageModels/Handlers/InputHandler.swift
```diff
@@ -88,3 +94,107 @@ public enum InputCoverage {
         }
     }
 }
+
+// MARK: - New InputHandler Protocol (inout fill pattern)
+
+/// Pre-allocated input buffer set. Owns NDArrays, reused across steps.
+/// The engine creates this once at init and passes it to `StaticInputHandler.fill()` each step.
+public struct InputBuffers {
```
carinapeng, line 106: `InputBuffers` still seems to expose COW cost? A future caller doing a `let a = buffers[name]` would reintroduce the +1 refcount

If they have to stay public, we could add warning of `get` not just on `_modify`

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
```diff
@@ -471,8 +494,21 @@ public final class StaticShapeEngine: InferenceEngine, @unchecked Sendable {
...
+
+        // Fill position_ids, causal_mask, step via the handler
+        let context = InputContext.static(
+            tokens: ArraySlice(batchTokens),
+            alignedStep: alignedStep,
+            batchSize: batchSize,
+            slidingWindow: nil,
+            contextBucket: contextLength)
+        try inputFiller.fill(context, into: &inputBuffers)
+        var inputs = inputBuffers.asDict()
```
carinapeng: `asDict()` copies the dict every step and the correctness leans on it not being retained?

swift/Sources/CoreAILanguageModels/Handlers/InputHandler+StaticBucket.swift
```diff
@@ -0,0 +1,124 @@
...
+        // Causal mask: [1, ctx, 1, batch] — lower-triangular
+        if let maskName = causalMaskName, let maskDesc = descs.causalMask {
+            buffers.ensureCapacity(name: maskName, descriptor: maskDesc)
+            try buffers.withMutableBuffer(maskName) { array in
+                array.mutableView(as: Float16.self)
+                    .withUnsafeMutablePointer { ptr, shape, strides in
+                        for ctx in 0..<shape[1] {
+                            for query in 0..<shape[3] {
+                                let offset = ctx &* strides[1] &+ query &* strides[3]
+                                ptr[offset] = Float16(-40000)
```
carinapeng: Minor: consider a named constant shared w the dynamic path so we don't drift between engines

swift/Sources/CoreAILanguageModels/Handlers/InputHandler+StaticBucket.swift
```diff
@@ -0,0 +1,124 @@
...
+                buffers.preAllocate(name: sName, descriptor: sDesc)
+            }
+        }
+    }
+
+    public func fill(_ context: InputContext, into buffers: inout InputBuffers) throws {
+        let batchSize = context.batchSize
+        let alignedStep = context.alignedStep
+        let tokensInBatch = context.tokens.count
+        let contextLength = context.contextBucket
```
carinapeng, line 74: Nit: we recompute `contextLength = context.contextBucket` and look up the bucket every step, we can make this more efficient by passing a resolved descriptor at graph selection time

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
stikves: Added `parseFunctionDimensions`

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAIStaticShapeEngine.swift
stikves: `resolveInputName(from:candidates:)` with explicit list

swift/Sources/CoreAILanguageModels/Handlers/InputHandler.swift
stikves: Added doc comments, we should not pollute output

carinapeng (approved):
Let's do more testing before merging :)

stikves:
Tested both correctness and performance

This new handler improves existing generation by 0.9% to 7.5% depending on the model and the chipset (by avoiding unnecessary copies)

Will merge after rebase.

PR #196, opened 2026-08-26
Add Qwen3-1.7B iOS and macOS export support

### LINKED ISSUE #116: Add Qwen3-1.7B support for iOS
### Source link

https://huggingface.co/Qwen/Qwen3-1.7B/tree/70d244cc86ccca08cf5af4e1e306ecf908b1ad5e

### Target platform

iOS

### Use case

Qwen3-1.7B fills the capability and memory gap between the existing Qwen3-0.6B and Qwen3-4B iOS options. It is useful for private on-device language tasks such as structured extraction, instruction following, short conversation, and retrieval over user-provided context without sending that content to a server.

### Preferred precision / compression

W8 per-tensor K-means-palettized transformer projections, separate INT8 per-tensor quantization for the tied embedding table, FP16 compute, and a fixed-size FP16 KV cache with a 4,096-token maximum context.

### Additional context

We prepared and validated a standalone Qwen3-1.7B iOS onboarding:

- Related community conversions exist, but they do not make this request a duplicate: one publishes a dynamic INT4 GPU bundle for `h18p` after its static ANE path failed to invoke, while another publishes a third-party INT8 macOS-path asset with no recorded AOT target and numeric accuracy reported as not run. This request is for an Apple-main preset, recipe, metadata, tests, and a validated static `h16p` path rather than another hosted compiled model.
- The preset is explicitly experimental and requires `--experimental`.
- The reference patch adds model metadata, the frozen compression recipe, registry entry, documentation, focused tests, and the iOS conversion matrix entry.
- The patch applies cleanly to Apple `coreai-models` main commit `04a3fd6cfe9bfae9cf05b1f246cf915d930d1c0a`.
- 28 focused unit tests and Ruff pass on that revision, and 42 related iOS conversion tests collect successfully.
- Clean export and `h16p` AOT compilation completed with 34 functions.
- An iPhone 15 Pro running iOS 27 completed four full six-case suites, including a 3,790-token long-context case and a reasoning case.
- The frozen W8 holdout produced mean cosine **0.996598**, minimum cosine **0.959625**, **98.37%** top-1 agreement, and mean NLL delta **0.003167**.
- Multiple W4 and mixed W4/W8 candidates were rejected before export, so the published W8 mechanism is the result of a frozen quality gate rather than a successful single-prompt smoke test.
- Apple Neural Engine participation was observed in a Core AI Instruments trace; exclusive ANE execution is not claimed.
- The [public reproduction repository](https://github.com/massif-01/qwen3-1.7b-coreai-ios) contains no model weights, tokenizer, compiled model, signed app, device identifier, local path, private prompt, or product source.

These measurements establish conversion fidelity for the frozen holdout, not a general downstream benchmark ranking. Full WikiText-2 perplexity is not included in the current evidence and is not represented as complete. The repository provides the recipe, reference patch, reproduction procedure, sanitized evidence, and resource-free companion app.

Related public artifacts, the ANE/GPU distinction, and the limits of cross-project comparison are documented in [`RELATED_WORK.md`](https://github.com/massif-01/qwen3-1.7b-coreai-ios/blob/main/RELATED_WORK.md).

stikves:
Register Qwen3-1.7B in the model registry for both macOS (4-bit) and iOS (mixed 4-bit/8-bit palettized) platforms. Add the mixed quantization config yaml and HuggingFace metadata entry.

Perplexity: 20.96 (float16) → 21.19 (mixed 4/8-bit, 5.43 BPW).

Closes #116


models/qwen3/README.md
```diff
@@ -7,6 +7,7 @@ Alibaba's Qwen3 models for on-device inference via Core AI.
 | Model      | Parameters | macOS | iOS |
 | ---------- | ---------- | ----- | --- |
 | Qwen3 0.6B | 0.6B       | Yes   | Yes |
+| Qwen3 1.7B | 1.7B       | Yes   | Yes |
```
tjia1818: let's make sure we tested AOT compilation + on-device specialization for this iOS variant.

models/qwen3/qwen3_1_7b_mixed_4bit_8bit.yaml
```diff
@@ -0,0 +1,19 @@
+kmeans_palettization_config:
+  global_config:
+    op_state_spec:
+      weight:
+        n_bits: 4
+        granularity:
+          type: per_grouped_channel
+          axis: 0
+          group_size: 8
```
pkmandke: if group_size=32 works, let's use that!

models/qwen3/README.md
stikves: Tested locally, yes it worked with expected speed and quality.

models/qwen3/qwen3_1_7b_mixed_4bit_8bit.yaml
stikves: It gives Infinity PPL. We are tracking that issue separately.

stikves:
Found a regression for iOS version locally. Moving to macOS only, and will follow up later

PR #200, opened 2026-08-26
enable activation quantization tests

pkmandke:
- Enable activation quantization tests that were previously disabled
- Add graph and eager mode parameterizations
- Remove try-catch in `compression.py` for `datasets` and `tqdm` since they're now in `pyproject.toml`

**Note:** The one pytest-skip is for the toy `Qwen2` in eager-mode and is being fixed in `coreai-opt` [here](https://github.com/apple/coreai-optimization/pull/78).


uv.lock
kevchengcodes, line 1: we need all these changes?

pkmandke: yeah seems like adding `"datasets>=4.0,<5.0"` to `pyproject.toml` pulls in all those..

pkmandke: I wonder if that was the reason why we're asking users to install datasets (and `tqdm`?) manually:

```py
if not _HAS_DATASETS:
        raise ImportError(
            "The 'datasets' and 'tqdm' packages are required for calibration data. "
            "Install them with: pip install datasets tqdm"
        )
```

I reverted it in this PR..

vineet-g (approved):

tjia1818 (approved):

uv.lock
stikves: Can we move the dependency from `dependencies` to a test one? Not everyone who just wants to export a model would want to pull all of these.

python/pyproject.toml
stikves, line 85: Can we add
```
[project.optional-dependencies.test]
```

So that the new dependency is only loaded for tests?

stikves (commented):
Looks good overall, please do the dependency move

python/pyproject.toml
pkmandke: Actually the new dependency (datasets) is in `compression.py` and will be used for calibration even by the export path not just tests..

PR #202, opened 2026-08-27
Add sliding window ring buffer cache for Muse Glimmer

stikves:
## Summary

Implement a compact KV cache for sliding-window attention (SWA) models that splits the cache
into fixed-size ring buffers (sliding layers) and growing caches (global layers). For Muse
Glimmer's 52-layer model (39 sliding + 13 global), this reduces KV memory from 7.0 GB to
1.83 GB at 128K context (74% savings).

**Primitives**:
- `RingKVCache`: fixed-size circular buffer, writes at `position % capacity`, never grows
- `ring_window_causal_mask`: position-aware mask for ring buffer attention

**Model changes** (muse_glimmer.py):
- 4 cache tensors (globalKey/Value + slidingKey/Value) with per-layer dense slot indexing
- Sliding layers use explicit ring mask; global layers use standard causal SDPA
- `export_state_classification()` for runtime state routing

**DFlash drafter** (muse_glimmer_drafter_ring.py):
- 5-layer all-SWA model using RingKVCache exclusively
- Shares embed_tokens + lm_head with 30B target (speculative decoding companion)
- Note: export/from_hf not wired yet (follow-up with speculative decoding PR)

**Swift runtime**:
- `isAllSlidingCache` detection in StateHandlerFactory
- Compact position_ids mode in CoreAISequentialEngine for ring buffer models

## Validation

- **PPL**: wikitext word_ppl = 7.7093 (baseline 7.71) on A100 via bolt eval
- **Parity**: incremental decode matches full recompute (max diff 7e-7)
- **Wrap-around**: 100 decode steps past window boundary, all finite and deterministic
- **17 unit tests**: 11 model structure + 6 ring buffer parity

## Known limitations (follow-up PRs)

- Drafter `from_hf` / `build_reference_inputs` / `build_dynamic_shapes` not implemented
- Swift StateHandlerFactory 2-state legacy path misclassifies drafter sliding states
  (needs `export_state_classification` wiring through metadata.json)
- Ring buffer `update_and_fetch` requires `write_start + query_len <= capacity`
  (safe for decode; chunked prefill must use half-window chunks)

## Test plan

- [x] Ruff lint + format clean
- [x] 17 pytest tests pass (model + parity)
- [x] Wikitext PPL matches baseline (7.71)
- [ ] Export through `coreai.llm.export` pipeline (requires follow-up wiring)
- [ ] End-to-end `llm-runner` generation (requires export)

### DIFF
--- a/python/src/coreai_models/export/pipeline.py
+++ b/python/src/coreai_models/export/pipeline.py
@@ -376,6 +376,7 @@ async def _async_export_model(config: ExportConfig) -> str:
             hf_config=hf_config,
             compression=config.compression,
             name=output_name,
+            tokenizer_model_id=entry.tokenizer_model_id,
         )
 
     logger.info(f"Export complete: {bundle_path}")
--- a/python/src/coreai_models/export/pipeline.py
+++ b/python/src/coreai_models/export/pipeline.py
@@ -75,6 +75,8 @@ class ExportConfig:
     compression_config_object: Any = field(default=None, repr=False)
     # Override HuggingFace model_type for registry lookup.
     model_type_override: str | None = None
+    # Speculative decoding: export the drafter model alongside the target.
+    with_drafter: bool = False
 
     def __post_init__(self) -> None:
         if self.quantization_mode == "graph" and self.variant != "macOS":
@@ -161,6 +163,15 @@ async def _async_export_model(config: ExportConfig) -> str:
 
     if config.variant == "iOS" and entry.ios_class is None:
         raise ValueError(f"Model '{model_type}' does not support iOS variant")
+
+    if config.with_drafter:
+        if config.variant != "macOS":
+            raise ValueError("--with-drafter is only supported for macOS variant.")
+        if entry.drafter_class is None or entry.drafter_model_id is None:
+            raise ValueError(
+                f"Model '{model_type}' does not have a registered drafter. "
+                "--with-drafter is not supported for this model."
+            )
 
     model_class = entry.macos_class if config.variant == "macOS" else entry.ios_class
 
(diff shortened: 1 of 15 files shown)


python/src/coreai_models/models/macos/muse_glimmer.py
```diff
@@ -333,21 +404,123 @@ def forward(
...
+                SLIDING_VALUE_CACHE_NAME,
+            )
+        }
+
+    @classmethod
+    def export_state_classification(cls) -> dict[str, str]:
+        return {
+            KEY_CACHE_NAME: "kv_cache",
+            VALUE_CACHE_NAME: "kv_cache",
+            SLIDING_KEY_CACHE_NAME: "sliding_cache",
```
srjoglekar246: Can we call this `sliding_kv_cache`... just to denote that its just the sliding version of the other "global" kv cache?

python/src/coreai_models/primitives/macos/cache.py
```diff
@@ -283,3 +283,124 @@ def update_states(
...
+        offset: int,
+        k: torch.Tensor,
+        v: torch.Tensor,
+        query_len: int | None = None,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
+        """Write K/V at ring position, return full cache for this layer.
+
+        The write must not wrap around the ring boundary, i.e.
+        ``(offset % capacity) + query_len <= capacity`` must hold.
+        Decode (query_len=1) always satisfies this. For chunked prefill,
```
srjoglekar246, line 358: Sometimes, the sliding window size could be small, like 512...and we have seen that for some models perf is maximized with a larger chunk size... in such cases, should we export a window of slightly larger size to optimize for the chunk?

python/src/coreai_models/export/pipeline.py
```diff
@@ -369,13 +380,60 @@ def get_calibration_data():  # type: ignore[no-untyped-def]
         # so offload it to a worker thread to keep the event loop responsive.
         metadata = build_aimodel_metadata(config.hf_model_id)
         await asyncio.to_thread(coreai_program.save_asset, aimodel_path, metadata)
+        del coreai_program
+
+        # ---- 6. Export drafter if requested ----
```
srjoglekar246, line 385: Would it be more optimal to export the drafter as part of the main function's output, to avoid materializing the `chunk_size x hidden_size` intermediates?

python/src/coreai_models/models/macos/muse_glimmer.py
stikves: Good call, updated

python/src/coreai_models/primitives/macos/cache.py
stikves: Chunk size is a runtime concept. I'm not sure we can handle it here.

Though we should have a mechanism in the swift runner to detect misalignment. I'll add that to follow up PR.

python/src/coreai_models/export/pipeline.py
stikves: I'm trying to get correctness first. The later iterations will optimize.

srjoglekar246 (approved):

PR #204, opened 2026-08-27
Optimize macOS LLMs prefill with distinct prompt function.

Lewis300:
# macOS: export a separate prefill graph and use it in the runner

macOS exported a single `main` graph serving both prefill and decode, so prefill paid for
the LM head on every prompt token even though only the last token's logits are sampled (or no tokens are sampled if prompt > chunk size).
This adds an optional second entrypoint and wires both engines to use it. iOS already
splits `extend`/`prompt_opt`; this does the same for macOS.

## Export

- **`prompt` entrypoint** beside `main` — same inputs and states, but no LM head and *no
  outputs*: it only fills the KV cache. One traced signature, traced twice; weights are
  shared between entrypoints.
- **Opt-in per model** via `BaseForCausalLM.exports_prompt_graph`, with `forward` honouring
  `prefill_mode`. All ten macOS text models are opted in; models that don't opt in export
  exactly as before. `Qwen3VLForCausalLMEmbeddings` is deliberately left out — it's exported
  by `vlm/export.py`, which doesn't ask for a prompt graph.
- **Graph mode is rejected at export.** A flattened `GraphModule` resolved `prefill_mode` at
  capture, so the flag would silently do nothing and the second trace would come out a copy
  of `main`. Failing at export beats failing at load.
    - **Follow up to support prompt graph in Graph mode:** Take the graph module (which includes the LM head) and cut out the logits output, leaving only the states.

## Runner

- Both the **pipelined and sequential** engines load `prompt` when the asset has it, run
  prefill chunks through it, and hold the final prompt token back for `main` — the prompt
  graph can't seed sampling. Lookup and validation are shared in `PromptGraph.swift`.
- Prefill now chunks **whenever a prompt graph is present** rather than only above
  `chunkThreshold`: with no LM head to pay for, there's no threshold to clear.
- The **logits buffer starts at one token** instead of 256 in that configuration, since
  `main` never sees more — hundreds of MB at large vocabularies that would go unused.
- Assets without a `prompt` graph take the old path unchanged.

## Testing

- **All macOS LLMs exporting and running. Runner picks up on prompt graph, output is sensible.**
- Four new tests in `test_macos_prompt_graph.py`: the graph is emitted only when opted in,
  it declares no outputs and binds the same inputs/states as `main`, its KV writes match
  eager torch prefill (fp32, atol/rtol 1e-2 against ~5e-3 observed error), and a flattened
  model is rejected. The parity test executes the graph, so it needs the local Core AI
  runtime; the other three don't.
- Greedy token IDs identical with and without the prompt graph on qwen2 and qwen3, across
  both engines, on 40- and 400-token prompts (the latter crossing several chunk seams).
- Full `test_export/` suite passes; `swift build` clean with no new warnings.


python/src/coreai_models/export/macos.py
```diff
@@ -20,6 +25,7 @@
 from coreai_models._constants import (
     DEFAULT_INCLUDE_DEBUG_INFO,
     MAIN_GRAPH_NAME,
+    PROMPT_GRAPH_NAME,
```
stikves: Let's use "PREFILL". This is not always a PROMPT.

Same everywhere else.

python/src/coreai_models/export/macos.py
```diff
@@ -54,10 +60,27 @@ def _build_reference_inputs(
     reference_inputs = model.build_reference_inputs(config, target_dtype, spec)
     dynamic_shapes = model.build_dynamic_shapes(config, spec)
     model.validate_export_contract(reference_inputs, dynamic_shapes)
-    # A macOS model has exactly one graph.
+    # A macOS model has exactly one traced signature; a `prompt` entrypoint, when the
```
stikves: for example here, and others.

python/src/coreai_models/export/macos.py
```diff
@@ -146,7 +185,19 @@ def export_fn(
...
                 "Call patch_model_for_externalization on the model before quantization."
             )
-        exported_program = export_fn(model, pass_inputs_as_kwargs=False)
+        if export_prompt_graph:
+            # The flattened graph resolved `prefill_mode` when it was captured, so the
+            # flag no longer does anything here and a second trace would come out a copy
+            # of `main` -- LM head, `logits` output and all, which the runner rejects.
+            # Emitting nothing is safe: it falls back to `main` for prefill.
+            logger.warning(
+                f"Not exporting the {PROMPT_GRAPH_NAME!r} entrypoint: this model arrives "
```
stikves: Isn't this an error?

python/src/coreai_models/models/macos/qwen3_moe.py
```diff
@@ -215,9 +218,14 @@ def forward(
         position_ids: torch.IntTensor,
         k_cache: torch.Tensor,
         v_cache: torch.Tensor,
-    ) -> torch.Tensor:
+    ) -> torch.Tensor | tuple:
         cache = KVCache(k_cache, v_cache)
         out = self.model(input_ids, position_ids, cache)
+        if self.prefill_mode:
+            # A bare `return` causes torch export to trace a leaf node with value
+            # `None` rather than having no leaf nodes whatsoever. Remedied with
```
stikves, line 226: This seems fragile

Let's find a way to handle the empty case for all models in the base class.

python/src/coreai_models/_constants.py
```diff
@@ -9,9 +9,15 @@
...
 
-# Graph name for a single-graph (macOS) export. iOS uses its own entrypoint names.
+# Graph name for the decode (and, absent a prompt graph, prefill) macOS export.
+# iOS uses its own entrypoint names.
 MAIN_GRAPH_NAME = "main"
 
+# Optional second macOS entrypoint: the prefill graph. Same inputs and states as `main`,
+# but no LM head and no outputs -- it only fills the KV cache. Emitted for models that opt
+# in via `BaseForCausalLM.exports_prompt_graph`; the Swift runner uses it when present.
+PROMPT_GRAPH_NAME = "prompt"
```
stikves: `"prefill"`

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -41,6 +41,10 @@ public final class CoreAISequentialEngine: InferenceEngine, @unchecked Sendable
     private let function: InferenceFunction
     private let functionDescriptor: InferenceFunctionDescriptor
 
+    // Optional prefill graph. Prefill chunks run here when the asset has it. It produces
+    // no logits, so the last prompt token still goes through `function`.
+    private let promptFunction: InferenceFunction?
```
stikves: `prefillFunction`

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
```diff
@@ -209,7 +219,9 @@ public final class CoreAISequentialEngine: InferenceEngine, @unchecked Sendable
     // MARK: - Prefill Strategy
 
     private func selectPrefillStrategy(newTokenCount: Int) -> PrefillStrategy {
-        if newTokenCount > config.chunkThreshold {
+        // With a prompt graph, chunking is cheaper at any size: every chunk but the last
+        // token skips the LM head, so there is no threshold to clear.
+        if promptFunction != nil || newTokenCount > config.chunkThreshold {
```
stikves: We might still want to keep this for memory optimization.

(Not sure any current model will immediately benefit, but one of the reasons of chunking is keeping smaller KV-caches when possible)

python/src/coreai_models/export/macos.py
Lewis300: Should we change iOS models to `prefill` as well instead of `prompt_opt`?

python/src/coreai_models/export/macos.py
Lewis300: See above^

python/src/coreai_models/export/macos.py
Lewis300: I will make a follow up to alter how we handle graph mode. I don't error-out here as we have tests which test graph mode that would just fail here

stikves (commented):
Thanks for the new functionality. Looks good overall

Please prefer `prefill` naming. And check other comments.

Also please add unit tests.

python/src/coreai_models/models/macos/qwen3_moe.py
Lewis300: We cannot exactly do this in the base class as individual model classes implement the forward method.

python/src/coreai_models/_constants.py
Lewis300: Should we change iOS models to `prefill` as well instead of `prompt_opt`?

Lewis300:
> Please prefer `prefill` naming. And check other comments.

I kept it as `prompt` to stay inline with how we export iOS prompt graphs with the name `prompt_opt`

> Also please add unit tests.

@stikves what is missing from the tests added in `python/tests/test_model_units/test_export/test_macos_prompt_graph.py`

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
Lewis300: For memory this is actually better as we will never materialize a logits tensor for anything larger than one token.

For TTFT performance there is some ambiguity regarding small prompts as two inference calls for a small prompt might be worse. For the case when prompt graph is present, the best option would be to have some token threshold. Where say, if there are less than 20 tokens to prefill, just execute the main graph rather than prefill 19 tokens in the prompt graph and the final token in the main graph.

python/tests/test_model_units/test_export/test_macos_prefill_graph.py
```diff
@@ -0,0 +1,276 @@
...
+    export_config = SimpleNamespace(max_context_length=MAX_CTX, compute_precision="float16")
+    with caplog.at_level(logging.WARNING, logger="coreai_models.export.macos"):
+        program = export_macos_model(flattened, config, export_config, externalized_model=model)
+
+    assert set(await _function_descriptors(program)) == {MAIN_GRAPH_NAME}
+    assert PREFILL_GRAPH_NAME in caplog.text
+
+
+@pytest.mark.asyncio
+async def test_prefill_graph_kv_writes_match_torch() -> None:
```
carinapeng, line 243: Thanks for adding this for numerical parity!

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
carinapeng, line 1: The pipelined engine warms multi token shapes on prefill graph, let's confirm sequential path does it too

So the first prefill doesn't eat the JIT cost

carinapeng (approved):
Thanks for adding the sequential engine wiring and tests for parity and Swift side as well

One note on the sequential engine prefill-graph warmup, and agree with Sukru's suggestion for some clean up

stikves (approved):
Thanks for adding the tests

Let's run models one last time before merging.

tjia1818 (approved):

python/src/coreai_models/export/macos.py
```diff
@@ -160,20 +211,41 @@ def export_fn(
...
+            )
     else:
         raise TypeError(
             "model must be a torch.nn.Module (eager-mode) or torch.fx.GraphModule "
             f"(graph-mode), got {type(model).__name__}."
         )
 
     register_custom_torch_lowering(converter)
-    return converter.to_coreai()
+    try:
```
vineet-g, line 243: Not sure if try/finally is needed. If the `converter.to_coreai()` step fails, model is useless anyways and export needs to be done again

Lewis300: This try block doesn't catch the error. It exists only to set the prefill mode back to the default (`False`) whether or not he conversion fails. In the event the user needs the base module again, it leave the object in the same state it was before attempting to export

Lewis300: This try block doesn't catch the error. It simply uses a `finally` block to re-set the prefill mode of the base module so that the caller of the function gets the model in the same state as it was when it was passed to this function, regardless of if to_coreai failed.

Lewis300: If to_coreai fails we will currently crash anyway. Just for the future in case we do catch the error somewhere above the model will not have its prefill-mode accidentally set to True inside the func

python/src/coreai_models/models/macos/qwen3_moe.py
Lewis300: We can't do that as each model class is responsible for implementing their own forward method.

Furthermore, in order to support prompt-graph separation when the user is quantizing in graph mode. I need a different approach of getting the prompt graph that renders this set-prefill-mode stuff irrelevant. It will be removed anyway.

pkmandke (approved):
+1 to @stikves comment- would be better if we can standardize and use one nomenclature prefill or prompt.

python/src/coreai_models/export/macos.py
Lewis300: No, we have tests that try to export a model when using graph mode quantization. I plan to make a follow up to support exporting a prompt graph when this is the case anyway. So no point in breaking the tests now.

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialEngine.swift
Lewis300: This if condition forces a chunked approach when we have a prefill function. This is because there is no point in pre-filling the entire batch with the prompt function, only to run extend on the final token once again since the prompt graph won't give us logits.

PR #211, opened 2026-08-29
Support prompt graph export for graph-mode quantized macOS models

Lewis300:
For models quantized using graph mode, the exporter gets a traced `GraphModule` instead of the original object. We cannot thus flip the `prefill_mode` of the object and export that as such branching logic is lost.

For `GraphModule`'s we instead trim all non-state outputs from the graph module to achieve the desired effect.


stikves (approved):

tjia1818 (approved):

pkmandke (commented):
Changes look good. Let's verify end to end with a few models, especially to ensure the graph pass works generically.

pkmandke (approved):

PR #221, opened 2026-09-02
Add iOS export preset for Qwen3-8B

chpriyal:
Qwen3-8B was registered for macOS only, so
`coreai.llm.export qwen3-8b --platform iOS` exited with "not available for iOS" even though the iOS model class (Qwen3ForCausalLMForiOS) already exists and is registered.

Register it for iOS using the default named palettization preset (4bit_weight_palettized_group32) at IOS_DEFAULT_MAX_CONTEXT_LENGTH, matching the style of the existing qwen2.5-1.5b-instruct iOS entry. No new compression YAMLs.

Adding this variant makes the short-name ambiguous, so include a regression test that a bare short-name still resolves to macOS via the macOS-preference branch in try_lookup_preset -- otherwise `coreai.llm.export qwen3-8b` would break.

The Mistral-7B iOS preset this change originally also added is now covered upstream at group size 8, so only the shared registry tests are extended to cover it.

### DIFF
--- a/models/qwen3/README.md
+++ b/models/qwen3/README.md
@@ -9,7 +9,7 @@ Alibaba's Qwen3 models for on-device inference via Core AI.
 | Qwen3 0.6B | 0.6B       | Yes   | Yes |
 | Qwen3 1.7B | 1.7B       | Yes   | Yes |
 | Qwen3 4B   | 4.0B       | Yes   | Yes |
-| Qwen3 8B   | 8.0B       | Yes   | No  |
+| Qwen3 8B   | 8.0B       | Yes   | Yes |
 
 ## Setup to export models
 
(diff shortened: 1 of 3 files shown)


models/qwen3/README.md
pkmandke, line 97: @chpriyal FYI: We'll need to add an entry for `iOS` PPL here once we have the numbers

python/src/coreai_models/model_registry.py
```diff
@@ -333,6 +333,16 @@ class UtilityModel:
...
     ),
+    ModelPreset(
+        "qwen3-8b",
+        "Qwen/Qwen3-8B",
+        "qwen3",
+        "llm",
+        "iOS",
+        "4bit_weight_palettized_group32",
+        "float16",
+        IOS_DEFAULT_MAX_CONTEXT_LENGTH,
```
stikves, line 344: Unrelated follow up: Shall we think about updating the `IOS_DEFAULT_MAX_CONTEXT_LENGTH` as well?

stikves (approved):

tjia1818:
We will need to check mixed precision and its ppl score to see if that is required. Closing for now.

PR #227, opened 2026-09-04
Add InputLayout for shared descriptor analysis; fix #212

### LINKED ISSUE #212: Pipelined engine cannot be constructed for bundles whose logits output has a static sequence length (GrowingLogitsBuffer initialCapacity 256 vs [1, 1, vocab])
Thanks for #103 — this is the same class of problem in the language-model path, and the same fix shape applies.

**Symptom.** `EngineFactory.createEngine` throws `Shape at dimension 1 of 256 is not a valid substitution for source shape 1` for any bundle whose `main` graph emits logits as a static `[1, 1, vocab]` (decode-only, S=1 graphs). It fails before any token is bound, so the prompt length does not matter. A third-party harness hit it on a public bundle: Liquid4All/pipette-clients#13 (`mlboydaisuke/Qwen3.8-27B-CoreAI`).

**Cause.** `CoreAIPipelinedEngine.swift:745-752` builds the logits buffer with `initialCapacity: averageExpectedPromptSize` (256, line 29), and `TensorStorage+CoreAI.swift:58-59` resolves `[1, capacity, vocabSize]` against the descriptor; a static dim-1 of 1 rejects 256.

**Fix (5 lines).** Size the buffer to the static length when the descriptor has one. This has been in production in the fork since June (john-rocky/coreai-models@9e5b605) on the community bundles published at huggingface.co/mlboydaisuke:

```swift
// A fully static logits output (e.g. a decode-only S=1 graph: [1, 1, vocab])
// can't be resolved at a larger capacity — size the buffer to its static
// sequence length instead of the prompt-sized default.
let logitsSeqIsStatic = logitsDesc.shape.count >= 2 && logitsDesc.shape[1] > 0
let logitsRef = try GrowingLogitsBuffer(
    device: device, descriptor: descriptor, name: logitsOutputName,
    vocabSize: config.vocabSize,
    maxCapacity: logitsSeqIsStatic ? logitsDesc.shape[1] : config.maxContextLength,
    initialCapacity: logitsSeqIsStatic ? logitsDesc.shape[1] : averageExpectedPromptSize
)
```

With that guard, `COREAI_CHUNK_THRESHOLD=1` (already honored by `ModelConfig.swift:126-142`) runs the multi-token prefill as S=1 steps on such bundles. A possible follow-up is to default the chunk size to 1 when `input_ids` is static S=1, so the env var is not needed; I left that out here to keep the change minimal.

Happy to open a PR with the guard (and a unit test against a static-logits descriptor) if that is easier to take than a patch in an issue.

stikves:
## Summary

- Adds `InputLayout`, a value type that centralizes descriptor name resolution and shape policy detection at engine init
- All three engines (pipelined, sequential, static shape) now use it instead of duplicating descriptor inspection
- Fixes #212: S=1 static-logits bundles (Qwen3.5 GDN) no longer crash the pipelined engine

**The #212 fix:** `GrowingLogitsBuffer` was sized to 256 against a static `[1,1,vocab]` descriptor. `InputLayout` detects the static logits dim and sizes the buffer correctly. It also auto-sets the prefill chunk threshold to the fixed query length, removing the need for `COREAI_CHUNK_THRESHOLD=1`.

## Changes

- `InputLayout.swift`: struct with `QueryPolicy`, `LogitsPolicy`, `PositionPolicy`, `PrefillPolicy` enums + `analyze()` factory
- `CoreAIPipelinedEngine.swift`: uses layout for logits buffer sizing and prefill chunk routing
- `CoreAISequentialEngine.swift`: uses layout for name resolution and position policy
- `CoreAIStaticShapeEngine.swift`: uses shared position_ids name resolution, removes dead `knownPositionIdNames`
- `InputLayoutTests.swift`: 19 tests covering all policy combinations and edge cases

## Test plan

- [x] 19 new InputLayout unit tests (name resolution, S=1 detection, prefill policy, edge cases)
- [x] Existing 430+ tests pass (only known flaky timing benchmarks fail)
- [ ] Verify with a real S=1 bundle (Qwen3.5 GDN) that pipelined engine constructs without crash
- [x] Verify dynamic model behavior is unchanged (same prefill routing, same logits buffer sizing)


tjia1818 (approved):

john-rocky:
Verified the merged commit `27a66f90e7f3fd9b83a6acb7bcb0a4a5ff71fd60` on real Qwen3.5 0.8B and Qwen3.8 27B GDN S=1 bundles. Both construct and generate 24 tokens with `COREAI_CHUNK_THRESHOLD` unset. The pre-fix source traps in the constructor; the dynamic Qwen3 0.6B control retains token-identical outputs across the #227 change.

Mac Studio M4 Max, macOS 27.0 `26A5416b`, Xcode 27 beta 5 `27A5237l`, SDK `26A5406c`. [Pinned bundle revisions, commands, hashes and full logs](https://github.com/john-rocky/coreai-models/blob/a40d26eaeb7dd5492c24edc178929e8e40daa184/validation/2026-09-05-coreai/RESULTS.md). The same official dependency also runs pipette's Swift HTTP contract on both S=1 bundles and its Rust CLI memory smoke through the normal sidecar build/cache path.

I isolated a separate last-sampled-token continuation issue and reported it in https://github.com/apple/coreai-models/issues/234, with a tested four-line patch. It also affects the pre-#227 dynamic control; I am not attributing it to this PR. These results cover the stated bundles and cases, not every multi-turn or review item.

PR #228, opened 2026-09-04
Add DFlash speculative decoding drafter for Muse Glimmer

stikves:
DFlash drafter with two-phase inference:
- inject_kv writes target encoder features into the ring KV cache via per-layer Wk/Wv projection.
- draft runs bidirectional attention over injected context and mask tokens, producing K token predictions in a single pass.

MuseGlimmerModelWithDrafter extracts hidden states at layers [1,13,25,37,49]. MuseGlimmerForCausalLMWithDrafter fuses encoder.fc + encoder_norm to produce (logits, drafter_features).

Parity verified against HF MuseGlimmerAssistantModel at all 24 sub-layer checkpoints. Acceptance: K=4 64%, K=8 30%, K=16 17%.


uv.lock
guru-desh, line 1: Why do we have so many uv lock changes?

stikves: removing, this was my local one

tjia1818 (approved):

PR #236, opened 2026-09-05
Diffusion pipeline latent preview

### LINKED ISSUE #231: PipelineProgress.currentLatent is never populated, so consumers cannot show a live preview
### What are you trying to build?

PipelineProgress has carried a currentLatent field since it was introduced, but Flux2Pipeline, SD3Pipeline and StableDiffusionPipeline all pass nil. Apps that want to show the image forming during generation have nothing to render.

Proposed change for FLUX.2, on my fork: https://github.com/apple/coreai-models/compare/main...james-333i:coreai-models:flux-preview-latents

After each Euler step it computes the flow-matching x0 prediction (x0 = x_t minus sigma times v), converts it to VAE latent space (unpack, batch-norm denorm, unpatchify) and hands it to the progress handler as a [1, 32, H/8, W/8] array. The final step is skipped since the decoded image follows immediately. The cost is a few array copies per step, negligible next to a transformer pass.

On the consumer side a cheap channel-to-RGB projection is enough for a recognizable preview. I fitted a 32x3 linear map against the FLUX.2 VAE by encoding a set of images and regressing against their downscaled RGB, and the previews show subject, color and composition clearly at every step. I can include those factors in a follow-up if a preview helper is wanted in the package itself. Glad to send this as a PR if PR creation opens up.

### Where are the current docs or utilities unclear?

N/A

### Expected improvement

Allows consumer to see progress on image generation steps.

### Additional context

_No response_

stikves:
## Diffusion pipeline latent preview

Surfaces intermediate denoising latents so a UI can render a lightweight live preview without a full VAE decode at every step.

- **`PipelineProgress.currentLatent`** — the current latent is exposed through an optional progress handler; when no handler is set, no preview work is performed.
- **`NDArray.asRGB(coefficients:)`** — projects a latent to an RGB image via a single `cblas_sgemm`. The projection is a small linear map specific to a model's VAE, supplied by the caller.
- **`LatentPreviewTuner`** and `diffusion-runner --tune-preview` / `--tune-fit` — collect latent/image pairs for a model, fit a coefficient set, and export per-step preview images.

Coefficients are fit per model rather than bundled, since the latent-to-RGB mapping depends on the model's VAE. See the `LatentRGBCoefficients` documentation for the tuning workflow.

Closes: https://github.com/apple/coreai-models/issues/231

### Follow-up (out of scope)

Preview resolution equals the latent resolution. A higher-fidelity option is a small distilled decoder that reconstructs full-resolution output at a fraction of the full VAE cost; tracked separately.

### Test plan

- [x] CI green
- [x] `--tune-preview` collects pairs; `--tune-fit` fits coefficients and exports previews
- [x] Live preview renders with a fitted coefficient set


.gitignore
```diff
@@ -34,3 +34,4 @@ python/.venv/
 # Local artifacts
 exports/
 *.jpeg
+output.png
```
tjia1818: can we just ignore *.png in general?

stikves: updated to ignore images in root folder

swift/Sources/CoreAIDiffusionPipeline/Pipelines/Flux2Pipeline.swift
```diff
@@ -423,8 +423,27 @@ public struct Flux2Pipeline: DiffusionPipeline {
             packedLatents = scheduler.step(output: output, timeStep: t, sample: packedLatents)
             try checkLatentsAreFinite(packedLatents, step: step)
 
-            let progress = PipelineProgress(step: step + 1, totalSteps: steps, currentLatent: nil)
-            if !progressHandler(progress) { break }
+            if let progressHandler {
+                // Unpack → denorm → unpatchify: [1, 128, 64, 64] → [1, 16, 128, 128]
```
kevchengcodes: I think these numbers would change based on the transformer function, can we make them more generalized? Also not certain about the 128 -> 16
```
[1, inChannels, spatialSide, spatialSide] → [1, inChannels / 4, spatialSide*2, spatialSide*2]
```

swift/Sources/CoreAIDiffusionPipeline/Pipelines/Flux2Pipeline.swift
```diff
@@ -423,8 +423,27 @@ public struct Flux2Pipeline: DiffusionPipeline {
...
+                // These are array copies, no model call.
+                let spatial = unpackLatentsSpatialFlatten(
+                    packedLatents, channels: inChannels, height: spatialSide, width: spatialSide)
+                let denormed = applyBatchNormDenorm(
+                    spatial, channels: inChannels, height: spatialSide, width: spatialSide)
+                let unpatchified = Self.unpatchifyLatents(
+                    denormed, channels: inChannels, height: spatialSide, width: spatialSide)
+
+                let vaeChannels = inChannels / 4  // 128 → 16 after patchify
```
kevchengcodes: is 128 -> 16 correct?

swift/Sources/CoreAIDiffusionPipeline/Pipelines/Pipeline.swift
```diff
@@ -47,8 +47,18 @@ public protocol DiffusionPipeline: ResourceManaging {
...
 }
+
+extension DiffusionPipeline {
+    /// Convenience: generate without a progress handler.
+    public func generateImages(
+        configuration: PipelineConfiguration
+    ) async throws -> GenerationResult {
+        try await generateImages(configuration: configuration, progressHandler: nil)
+    }
+}
```
kevchengcodes, line 64: cleaner option might be to make progressHandler=nil as default in the signature so we don't need an extension. Up to you though

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreview.swift
```diff
@@ -0,0 +1,115 @@
...
+    ///
+    /// `coefficients` must match the model's VAE; fit a set via
+    /// `diffusion-runner --tune-preview` / `--tune-fit`.
+    public func asRGB(
+        coefficients: LatentRGBCoefficients
+    ) -> CGImage? {
+        return draftProjection(coefficients: coefficients)
+    }
+
+    /// Project C-channel latent to 3-channel RGB using vDSP matrix multiply.
```
kevchengcodes: I don't see it using vDSP, should this say `cblas_sgemm` or BLAS instead?

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreview.swift
```diff
@@ -0,0 +1,115 @@
...
+        view.withUnsafePointer { ptr, _, _ in
+            for c in 0..<channels {
+                for p in 0..<spatialCount {
+                    nxc[p * channels + c] = ptr[c * spatialCount + p]
+                }
+            }
+        }
+
+        // GEMM: [N, C] × [C, 3] → [N, 3]
+        var nx3 = [Float](repeating: 0, count: spatialCount * 3)
```
kevchengcodes: I'm assuming this will always be 3, but can we use a defined coefficient for this

swift/Sources/Tools/diffusion-runner/DiffusionRunnerMain.swift
```diff
@@ -23,7 +23,7 @@ struct DiffusionRunner: AsyncParsableCommand {
     )
 
     @Option(help: "Path to model directory containing .aimodel components (or pipeline.json)")
-    var model: String
+    var model: String?
```
kevchengcodes, line 26: why is this optional now?

.gitignore
```diff
@@ -33,4 +33,5 @@ python/.venv/
 
 # Local artifacts
 exports/
-*.jpeg
+/*.jpeg
+/*.png
```
kevchengcodes, line 37:
```suggestion
*.jpeg
*.png
```

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+import Accelerate
+import CoreAI
+import CoreAIShared
+import CoreGraphics
+import Foundation
+import ImageIO
+
+/// Fits a linear latent-to-RGB projection for real-time diffusion previews.
+///
+/// The tuning workflow has two phases:
```
kevchengcodes, line 15: We should either put these instructions somewhere more visible or have READMEs point to it

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+            let image = decodedImage
+        else {
+            print("[tune] fit failed: latents=\(latents.count), hasImage=\(decodedImage != nil)")
+            return nil
+        }
+
+        let shape = lastLatent.shape
+        print("[tune] lastLatent shape=\(shape), data.count=\(lastLatent.data.count)")
+        guard shape.count == 4, shape[0] == 1 else {
+            print("[tune] fit failed: shape rank \(shape.count) != 4")
```
kevchengcodes: the print doesn't include this guarded expression: `shape[0] == 1`

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+
+            // Full (VAE) preview
+            if let vaeDecode {
+                do {
+                    let decoded = try await vaeDecode(data, shape)
+                    savePNG(decoded, to: outputDir.appendingPathComponent("latent_\(step)_vae.png"))
+                } catch {
+                    print("[tune] VAE decode failed at step \(step): \(error)")
+                }
+            }
```
kevchengcodes: is there a point to doing the VAE decode as part of the preview? Won't we be doing this anyway after the denoiser completes

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+        }
+        let modes = vaeDecode != nil ? "draft + VAE" : "draft only"
+        print("[tune] Exported \(latents.count) step previews (\(modes))")
+    }
+
+    private func savePNG(_ image: CGImage, to url: URL) {
+        Self.writePNG(image, to: url)
+    }
```
kevchengcodes: feels like a super thin wrapper we can remove

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+        guard let dest = CGImageDestinationCreateWithURL(url as CFURL, "public.png" as CFString, 1, nil) else {
+            return
+        }
+        CGImageDestinationAddImage(dest, image, nil)
+        CGImageDestinationFinalize(dest)
+    }
+
+    /// Downsample a CGImage to the target resolution and extract interleaved RGB bytes.
+    private static func extractRGB(from image: CGImage, width: Int, height: Int) -> [UInt8]? {
+        let bytesPerRow = width * 4
```
kevchengcodes, line 284: is this math always true?

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+        )
+
+        // Solve via LAPACK sposv_ (symmetric positive definite)
+        var n = Int32(channels)
+        var lda = n
+        var ldb = n
+        var nrhs = Int32(3)
+        var info = Int32(0)
+        var uplo = Int8(UInt8(ascii: "U"))
+        sposv_(&uplo, &n, &nrhs, &xtx, &lda, &xty, &ldb, &info)
```
kevchengcodes: there's a lot of duplicate code between `fitFromDirectory` and `fitCoefficients`. Do we need both paths, and if so, can we unify the math operations into a single helper?

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
```diff
@@ -0,0 +1,603 @@
...
+                .sorted {
+                    extractStepNumber(from: $0.lastPathComponent) < extractStepNumber(from: $1.lastPathComponent)
+                }
+
+            for file in latentFiles {
+                guard let (data, shape) = readLatentBin(file) else {
+                    print("[tune-fit] Could not read \(file.lastPathComponent), skipping")
+                    continue
+                }
+                guard shape.count == 4, shape[0] == 1 else { continue }
```
kevchengcodes: could add a print here, unless you feel it's redundant

.gitignore
stikves: We don't want to ignore potential images in other folders like docs, this is to prevent accidental local outputs.

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
stikves: We don't have a readme, we can move it when we have a docs folder. For now the code comments should be sufficient.

swift/Sources/CoreAIDiffusionPipeline/Pipelines/Pipeline.swift
stikves: Can't have a default in a protocol.

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreview.swift
stikves: We always have 3 channels (RGB), it is extremely unlikely this will change.

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
stikves: This is for visual verification.

swift/Sources/Tools/diffusion-runner/DiffusionRunnerMain.swift
stikves: Yes, the tuning from preexisting directory of outputs does not need a model.

swift/Sources/CoreAIDiffusionPipeline/Pipelines/Flux2Pipeline.swift
stikves: The comment is typo (will update to 32), but the code is already correct, and the logic is from existing unpatchify (assume the comment is a representative example, the code is already generic)

In any case, it might make sense to refactor this to a single helper.

swift/Sources/CoreAIDiffusionPipeline/Pipelines/Flux2Pipeline.swift
stikves: will update

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
stikves: Let me see what we can safely merge together

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreviewTuner.swift
stikves: It is true by construction, we pass 4 to the init below

swift/Sources/CoreAIDiffusionPipeline/Preview/LatentPreview.swift
stikves: nice catch. Both come from Accelerate, but let's be more exact.

swift/Sources/Tools/diffusion-runner/DiffusionRunnerMain.swift
kevchengcodes: Ah I see. Can we add a guard that model must be passed if the tuning options aren't passed?

kevchengcodes (approved):
looks good, left one more comment

swift/Sources/Tools/diffusion-runner/DiffusionRunnerMain.swift
stikves: Yes, we have that below:

```swift
        guard let model else {
            print("Error: --model is required for generation")
            throw ExitCode.failure
        }
```

PR #239, opened 2026-09-05
Fix SD 1.5 NaN output: float32 compute precision and scheduler lower-order final

stikves:
Two independent fixes for blank/NaN output from SD 1.5 and SD 2.1:

1. Change compute_precision from float16 to float32 for SD 1.5 and SD 2.1 exports. Float16 attention Q*K^T products overflow fp16 max (65504) during later denoising steps, producing Inf then NaN after softmax.

2. Remove the timeSteps.count < 15 gate from DPM-Solver++ lowerOrderFinal. HuggingFace diffusers applies lower_order_final unconditionally. With 20 steps, the second-order solver on the final step amplifies errors via 1/r0 (~5x), which compounds the fp16 overflow.

### DIFF
--- a/swift/Sources/CoreAIDiffusionPipeline/Components/CoreAIDiffusionModelFunction.swift
+++ b/swift/Sources/CoreAIDiffusionPipeline/Components/CoreAIDiffusionModelFunction.swift
@@ -402,6 +402,7 @@ public enum CoreAIDiffusionError: Error, LocalizedError {
     case unsupportedOutputScalarType(NDArray.ScalarType)
     case expectedSingleOutput(got: [String])
     case inputCountMismatch(name: String, shape: [Int], expected: Int, got: Int)
+    case nanDetected(step: Int)
 
     public var errorDescription: String? {
         switch self {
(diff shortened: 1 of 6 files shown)


swift/Sources/CoreAIDiffusionPipeline/Components/CoreAIDiffusionModelFunction.swift
```diff
@@ -419,6 +421,19 @@ public enum CoreAIDiffusionError: Error, LocalizedError {
...
         }
     }
 }
+
+// MARK: - NaN Detection
+
+func checkLatentsForNaN(_ latents: [Float], step: Int) throws {
+    var sum: Float = 0
+    vDSP_sve(latents, 1, &sum, vDSP_Length(latents.count))
+    if sum.isNaN || sum.isInfinite {
```
tjia1818: is `.isInfinite` intended? the error code does not reflect that

stikves: Good catch. We can call this `isFinite` instead, and cover both cases.

tjia1818 (approved):

PR #240, opened 2026-09-07
Layered prefill chunk size

stikves:
## Summary

Replace the single `COREAI_CHUNK_THRESHOLD` environment variable with a layered override system for prefill chunking, exposing two independent settings:

- **`prefillChunkSize`** — tokens processed per prefill chunk.
- **`prefillChunkThreshold`** — minimum prompt length (in tokens) that triggers chunking; prompts at or below it are processed in a single pass.

## Resolution order

Each setting is resolved by the first source that supplies a value, highest precedence first:

1. CLI flag — `--chunk-size` / `--chunk-threshold`
2. Foundation Models API — `CoreAILanguageModel(prefillChunkSize:prefillChunkThreshold:)`
3. `metadata.json` — `prefill_chunk_size` / `prefill_chunk_threshold`
4. Deprecated `COREAI_CHUNK_THRESHOLD` environment variable (emits a one-time warning)
5. Memory-based default

This precedence applies uniformly across the CLI tools (`llm-runner`, `llm-server`, `benchmark`), the Foundation Models API, and both the text and vision-language paths.

## Memory-based default

The default chunk size scales with installed RAM:

| Installed RAM | Default chunk size |
| ------------- | ------------------ |
| ≤ 24 GB       | 2048 tokens        |
| > 24 GB       | 4096 tokens        |

The chunk threshold defaults to 2× the resolved chunk size.

This default applies to the dynamic (GPU) engine. The static-shape (Neural Engine) path derives its chunk width from the compiled model's fixed query-length buckets and does not consult this value.

## Deprecation

`COREAI_CHUNK_THRESHOLD` continues to work and emits a one-time deprecation warning. Its value is applied as the chunk size (with the threshold set to the same value). Prefer `--chunk-size` or the `prefill_chunk_size` metadata key.

## Test plan

- `ChunkingConfigTests` covers layered resolution (override > env var > default), `metadata.json` decoding, invalid/zero overrides, Codable round-trip (overrides are runtime-only), the deprecated env var, and propagation through the VLM config.
- `CoreAILanguageModels`, `llm-runner`, `llm-server`, and `benchmark` build clean.


swift/Sources/CoreAILanguageModels/InferenceEngines/ModelConfig.swift
```diff
@@ -123,23 +135,61 @@ extension ModelConfig {
...
+            return override
+        }
         if let value = ProcessInfo.processInfo.environment["COREAI_CHUNK_THRESHOLD"],
             let size = Int(value), size > 0
         {
+            Self._emitDeprecationWarning()
             return size
         }
-        return 1024
+        return defaultPrefillChunkSize()
```
kevchengcodes, line 153: Looks like `defaultPrefillChunkSize()` will either choose 2048 or 4096. But this used to return 1024 if it reached this point. Have we tested performance knowing this?

swift/Sources/CoreAILanguageModels/InferenceEngines/ModelConfig.swift
```diff
@@ -123,23 +135,61 @@ extension ModelConfig {
...
+    /// Prefill chunk size with layered resolution:
+    /// override → deprecated env var → memory-based default.
+    public var prefillChunkSize: Int {
+        if let override = prefillChunkSizeOverride, override > 0 {
+            return override
+        }
         if let value = ProcessInfo.processInfo.environment["COREAI_CHUNK_THRESHOLD"],
             let size = Int(value), size > 0
         {
+            Self._emitDeprecationWarning()
```
kevchengcodes, line 150: should we emit warning or remove using the environment variable entirely?

swift/Sources/CoreAILanguageModels/InferenceEngines/ModelConfig.swift
```diff
@@ -123,23 +135,61 @@ extension ModelConfig {
...
-        return min(512, chunkThreshold)
+    /// Chunk threshold with layered resolution:
+    /// override → default (2× prefillChunkSize).
+    /// When the deprecated env var is active, threshold equals the env var value.
+    public var prefillChunkThreshold: Int {
+        if let override = prefillChunkThresholdOverride, override > 0 {
+            return override
+        }
+        if ProcessInfo.processInfo.environment["COREAI_CHUNK_THRESHOLD"] != nil {
+            return prefillChunkSize
```
kevchengcodes, line 164: if we do decide to stick with just emit warnings, there is no emit warning here

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
```diff
@@ -151,7 +151,15 @@ public final class CoreAISequentialVLMEngine: MultimodalInferenceEngine, @unchec
         llmModel: PreparedModel,
         options: EngineOptions = EngineOptions()
     ) async throws {
-        self.config = config
+        // Apply runtime chunking overrides (CLI / metadata) onto the base config,
+        // mirroring EngineFactory.selectEngine for the standard LLM path. `base` is
+        // immutable, so resolve on a local copy and reconstruct the VLM config.
+        var resolvedBase = config.base
+        resolvedBase.applyChunkingOverrides(
```
kevchengcodes, line 158: just a note for a future PR: should `CoreAISequentialVLMEngine` be integrated into `EngineFactory`? Right now we're duplicating some code here because they are separate

swift/Sources/Tools/llm-server/LLMServerMain.swift
```diff
@@ -51,6 +51,18 @@ struct LLMServer: AsyncParsableCommand {
     @Option(name: .customLong("kv-cache-initial-capacity"), help: "Initial KV cache capacity in tokens")
     var kvCacheInitialCapacity: Int?
 
+    @Option(
+        name: .customLong("chunk-size"),
+        help: ArgumentHelp("Prefill chunk size in tokens (default: memory-based, use 128 for MoE)", visibility: .hidden)
```
tjia1818: "use 128 for MoE" is a suggestion or default behavior? Let's be clear on that.

swift/Sources/CoreAILanguageModels/InferenceEngines/ModelConfig.swift
stikves: Yes, this is based on perf team feedback.

swift/Sources/CoreAILanguageModels/InferenceEngines/ModelConfig.swift
stikves: There might be existing scripts that depends on this.

We will eventually remove it

swift/Sources/CoreAILanguageModels/InferenceEngines/CoreAISequentialVLMEngine.swift
stikves: This is on the long term TODO list, there is still some more work before they can merge

swift/Sources/Tools/llm-server/LLMServerMain.swift
stikves: It is a clear suggestion. Based on perf tests

But we can update the language

tjia1818 (approved):

PR #243, opened 2026-09-09
Export Script for SAM3 Video Segmentation

kevchengcodes:
## Summary

This PR introduces the SAM3-Video Core AI export recipe. A future PR will bring in the Swift runtime to segment a video using this export.

The export matches the Hugging Face (HF) implementation as closely as possible. However, the HF `Sam3VideoInferenceSession` cannot be converted/traced in the Core AI export. Therefore, we  will extract that into the Swift runtime and split the Python export into separate modules for the image encoding, text encoding, detection, and tracking/memory bank functions.

## Usage

A new model card and export entry script can be found at `models/sam3_video`.

```bash
uv run models/sam3_video/export.py
```

Note that under the hood, this script uses the same `coreai.segmentation.export` module, but routes to the `--video` export pipeline instead. The existing image segmentation export functionality is maintained.

```bash
# for image segmentation, use the same commands as before
uv run models/sam3/export.py # for the lite model intended for iOS
uv run models/sam3/export.py --full # for the vanilla HF model
```


models/sam3_video/README.md
```diff
@@ -0,0 +1,123 @@
+# SAM 3 Video Segmentation
+
+SAM 3 detects objects matching a text prompt on every frame of a video and links them
```
tjia1818, line 3: segments or detects?

stikves (approved):

models/sam3_video/README.md
kevchengcodes: In this pipeline, I think detects is more accurate

PR #244, opened 2026-09-10
Fix pipelined engine crash with repetition penalty during prefill

stikves:
## Problem

Running `--repetition-penalty` on the pipelined engine crashes during prefill:

```
MPSGraphExecutable.mm: failed assertion 'Incompatible element type for
parameter at index 1, mlir module expected element type f16 but received f32'
```

When `repetitionPenalty` is set, `MPSGraphCompositeSampler` compiles the graph
with **6 feeds**: logits, penalty, temperature, random, topP, minP. But the
non-penalty `encode()` and `encodeWithSlice()` paths — used during prefill and
for the argmax fast path — passed only **5 inputs**, omitting the penalty tensor.
The f32 temperature data landed where the f16 penalty tensor was expected.

The decode path worked (it uses the penalty-aware `encode()`), so the crash only
surfaced during prefill — after `#176` shipped repetition penalty.

## Fix

- Add a pre-allocated **neutral penalty buffer** (all `1.0` in f16) allocated
  when `penaltyEnabled`.
- Add a `buildInputs()` helper that orders inputs by `executable.feedTensors`
  operation names, supplying the neutral buffer when the caller has no penalty
  data. All three encode paths now feed the correct 6 tensors.

## Verification

```
llm-runner --model qwen3_4b_4bit_dynamic --repetition-penalty 1.3 --temperature 0.8
```

Now generates coherent output during both prefill and decode (97 tok/s, no crash).


tjia1818 (approved):

tjia1818:
any unit we can add to prevent this?

stikves:
Let me double check something, this could be a local issue on my branch.

stikves:
Verified on tip of main

```sh
[561 / 561] llm-runner-product

Build complete! (78.79 sec)

⏳ Preparing AI asset from source... done in 9.108s

Generating...
MPSGraphExecutable.mm:4490: failed assertion `Incompatible element type for parameter at index 1, mlir module expected element type f16 but received f32'
zsh: abort      swift run -c release llm-runner --model exports/qwen3_4b_4bit_dynamic  1.3 
```

The fix is necessary

PR #246, opened 2026-09-11
Fix Phi-4 RoPE runtime regression while preserving LongRoPE

stikves:
## Summary

The runtime composite RoPE op rejects `freqs.count != head_dim/2` on newer OS builds, which forces partial-rotary models (Phi-4-mini, `partial_rotary_factor=0.75`) onto a slow CPU scalar fallback (~2.7 tok/s instead of ~100 tok/s).

This routes partial-rotary models through the **decomposed** (raw torch ops) RoPE path, which the runtime lowers correctly.

## Why not just use plain `DecomposedRoPE`

A plain `DecomposedRoPE(dims, base)` emits *unscaled* rotary and silently drops Phi-4's LongRoPE scaling — the `attention_factor` (~1.19 for the 131072/4096 ratio) and the per-dimension short/long factors. Measured against the HF-faithful `LongRoPE`:

| context | attention-logit mean rel. error (plain decomposed) |
|---|---|
| 16 / 512 | ~29% (from `attention_factor` alone) |
| 4096 | 600%+ (per-dim factor mismatch compounds with position) |

So the decomposed path must **carry the LongRoPE scaling**:

- `DecomposedRoPE` gains optional `inv_freq` + `attention_scale`.
- `LongRoPE.to_decomposed()` reuses its (unit-tested, HF-validated) frequencies and attention factor.
- `initialize_rope(..., decomposed=True)` returns the correct decomposed equivalent per rope type, and unifies the existing `FORCE_DECOMPOSED_ROPE` env path.
- `phi3.py` just passes `decomposed=partial_rotary_factor < 1.0`.

## Verification

- `LongRoPE.to_decomposed()` output is **bitwise-equal (0.00)** to `LongRoPE` at seq 16/512/4096.
- New `TestDecomposedPartialRotary` parity tests (fail on the plain path, pass here).
- Full `test_phi3.py` + macOS `test_rope.py` pass (40 tests); ruff clean; no new type errors.
- Full-rotary models (Phi-3/3.5) are unaffected — they keep the composite RoPE path.

### DIFF
--- a/python/src/coreai_models/primitives/macos/rope.py
+++ b/python/src/coreai_models/primitives/macos/rope.py
@@ -46,16 +46,27 @@ class DecomposedRoPE(torch.nn.Module):
       y1 = cos * x1 - sin * x2
       y2 = sin * x1 + cos * x2
       output = cat(y1, y2, passthrough)
+
+    ``inv_freq`` and ``attention_scale`` let this reproduce a scaled RoPE
+    variant (e.g. LongRoPE) with raw ops: pass precomputed per-dimension
+    inverse frequencies and the attention factor that scales the rotary
+    dims before rotation. See :meth:`LongRoPE.to_decomposed`.
     """
 
     def __init__(
         self: Self,
         dims: int | None = None,
         base: float = 1e4,
+        inv_freq: torch.Tensor | None = None,
+        attention_scale: float = 1.0,
     ) -> None:
         super().__init__()
         self.dims = dims
         self.base = base
+        # Stored as a plain attribute (like LongRoPE._freqs) so it stays a
+        # constant under torch.export; moved onto the input device in forward.
+        self._inv_freq = inv_freq
+        self.attention_scale = attention_scale
 
     def forward(
         self: Self,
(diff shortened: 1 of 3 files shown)


python/src/coreai_models/primitives/macos/rope.py
```diff
@@ -271,6 +292,19 @@ def forward(
             offset=offset,
         )
 
+    def to_decomposed(self: Self) -> "DecomposedRoPE":
```
carinapeng, line 299: We are silently dropping `interleaved` here, like `LongRoPE._rope = RoPE(..., interleaved=interleaved)`. Even though DecomposedRoPE always uses the split-half convention there's no reason that someone wouldn't wire up an interleaved model with this, because you are putting this in the primitives rope.py

Suggest either carry it through or assert this assumption

python/src/coreai_models/primitives/macos/rope.py
```diff
@@ -353,4 +389,12 @@ def initialize_rope(
             msg = f"Unsupported RoPE type {rope_type}"
             raise ValueError(msg)
 
+    if use_decomposed:
```
carinapeng, line 399: Now, anything that isn't long rope or default rope type is raising NotImplementedError, but phi3 sets `decomposed=partial_rotary_factor < 1.0`

So a partial rotary model configured with ie linear or dynamic scaling which previously would export (slow) composite path now is failing

Let's confirm we don't have any partial rotary models that use non default or non LongRoPE  types. We could also make the error name the model/config or fall back to composite for unsupported types

carinapeng (approved):
This looks right to me, but two follow ups before merging to consider, thank you!

python/src/coreai_models/primitives/macos/rope.py
stikves: Added a guard

PR #250, opened 2026-09-13
Add async request queue to llm-server

stikves:
## Summary

The llm-server currently gates concurrency with a binary busy flag: while one
request is generating, any concurrent request is rejected immediately with
`429`. This PR replaces that gate with an async `RequestQueue`, so concurrent
requests wait in a bounded queue and are rejected only when the queue is full.

## Changes

- **`RequestQueue`** (`CoreAILMCommon`): an async semaphore built on
  `withCheckedThrowingContinuation` with an acquired / rejected / queued state
  machine. Replaces `tryAcquire()`/`release()` and the `generating` flag.
  `acquire()` returns a **`QueuePermit`** that frees the slot exactly once — on
  explicit `release()` or, as a backstop, on deallocation, so a dropped
  streaming response can never leak the slot and wedge the server.
- Queued waiters are **cancellation-aware**: if the awaiting task is cancelled
  while queued (e.g. on shutdown), the waiter is removed and its `acquire()`
  throws `CancellationError` instead of leaking its continuation or being handed
  the slot after the fact.
- **`--max-queue-depth`** (default 16): how many requests may wait before the
  server returns `429` with a `queue_full` error. Rejected as invalid if
  negative; `0` means no queuing (second concurrent request gets an immediate
  `429`).
- **`hasRecurrentState`** on `InferenceEngine` (default `false`; overridden by
  the sequential and pipelined engines): recurrent (SSM / hybrid) models can't
  reuse a token prefix, so the server skips the prefix-reuse fast path and counts
  a miss. Engine correctness on rewind is already enforced by the engine itself;
  this only keeps the server's reuse stats and logs accurate.
- **`ServerError.queueFull(depth:)`** for the rejection path.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Single request | Served | Served (unchanged) |
| Concurrent requests | Immediate `429` | Queued, served FIFO |
| Queue at capacity | Immediate `429` | `429` (`queue_full`) |

`/ready` now reports busy based on queue activity rather than the removed
`generating` flag.

## Testing

- `RequestQueue` unit coverage (in `CoreAILMCommonTests`): immediate acquire,
  concurrent waiting, FIFO wakeup, rejection at `maxDepth`, `maxDepth == 0`,
  negative-depth clamp, idempotent permit release, and cancellation of a queued
  waiter. `CoreAILMCommonTests` (52) pass; `llm-server` builds.
- Load-tested end to end against `muse_glimmer_30b_4bit_dynamic`:
  - Single request: served.
  - 8 concurrent (< depth): all served FIFO, no premature `429`.
  - 30 concurrent (> capacity): 17 served, 13 rejected with `queue_full`. This
    matches 1 in-flight + 16 queued at the default depth.
  - 20-request soak: no errors; server returned to idle with no stuck slots.


swift/Sources/CoreAILMCommon/RequestQueue.swift
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// MARK: - Queue Permit
+
+/// A held `RequestQueue` slot. Releases exactly once — on the first `release()`
+/// call or, as a safety net, on deallocation (e.g. if a streaming response body
+/// is dropped by the server before its writer closure ever runs).
+public final class QueuePermit: Sendable {
+    private let queue: RequestQueue
+    private let released = Mutex<Bool>(false)
```
alejandro-isaza: Consider Atomic instead.

alejandro-isaza (approved):

swift/Sources/CoreAILMCommon/RequestQueue.swift
stikves: Makes sense, I'll update.

PR #251, opened 2026-09-15
Consolidate VLM sequential engine: shared iterator helpers, tidy (B3+B4)

stikves:
Follows #249 (B1+B2). Third step of the VLM/text sequential-engine consolidation.

## Changes
- B3: move the max-token clamp and next-token selection (identical in `CoreAISequentialEngine` and `CoreAISequentialVLMEngine`) into a `SequentialIterator` namespace both call. Each engine keeps its own `next()` loop: the text engine's batched forced-continuation PPL path and prefix cache, the VLM engine's image prefill.
- B4: drop the addressed re-use TODO in the VLM engine, fold the repeated reset/warmup state-clear into `clearGenerationState()`, update the class doc.

No behavior change.

## Tests
New `SequentialIteratorTests` (clamp incl. clamp-down and zero-floor, next-token). Existing suites pass under `swift test --disable-sandbox`.

## Validation
Bit-exact vs the pre-refactor baseline: text greedy 128/128, forced-continuation PPL exact; VLM greedy 115/115 and 400/400.

## Notes
This is steps 3+4 of 6:

This is a clean refactor of common core, the next one will do modifications on the iterator.


swift/Sources/CoreAILanguageModels/InferenceEngines/SequentialIterator.swift
```diff
@@ -0,0 +1,45 @@
...
+        }
+        return min(requested ?? Int.max, max(0, maxContextLength - inputCount))
+    }
+
+    /// Select the next token: the forced-continuation token when replaying, otherwise the
+    /// sampler's choice.
+    ///
+    /// `logits` is taken by value; the sampler mutates a copy-on-write copy so the caller's
+    /// buffer (which it may also return to the consumer) is left untouched — matching the
+    /// engines' previous `var mutableLogits = logitBuffer` behavior.
```
alejandro-isaza: "matching the engines' previous `var mutableLogits = logitBuffer` behavior." is superfluous, remove

alejandro-isaza (approved):

PR #255, opened 2026-09-15
SAM3 Video Runtime

kevchengcodes:
## Summary

This commit ports over the SAM3 video pipeline from the Transformers Python library into a new `CoreAIVideoSegmenter` Swift library. The majority of the code is purely additive and works on top of the existing SAM3-Video export.

### Changes
- new `video-segmenter` Swift CLI tool for quick inference and debugging
  - includes parity infrastructure for parity checks against BYO references
 - new 'CoreAIVideoSegmenter` Swift target containing the video pipeline
   - also includes utilities to render mask/box/label results over the input video
 - several new utilities in `CoreAIShared`

## How to Use
Please see `models/sam3_video/README.md`


models/sam3_video/README.md
tjia1818: looks this can be named as `--input` or `--input-video` to be more intuitive. We may also want to rename that for image-segmenter CLI.

tjia1818: you meant 15th frame here?

swift/Sources/CoreAIVideoSegmenter/Session/MemoryBankPacker.swift
```diff
@@ -0,0 +1,321 @@
...
+import CoreAI
+import CoreAIShared
+import Foundation
+
+/// The seven fixed-shape memory tensors `tracker_step` reads, reused across calls.
+struct PackedMemory {
+    var spatialMemory: NDArray
+    var spatialMemoryPosition: NDArray
+    var spatialTemporalIndex: NDArray
+    var spatialValid: NDArray
```
tjia1818: this var name a bit odd (ends with an adjective), same as `pointerValid`.

swift/Sources/CoreAIShared/Bundle/BundleKind.swift
```diff
@@ -13,5 +13,8 @@ public enum BundleKind: String, Codable, Sendable, CaseIterable {
     case vlm
     case diffusion
     case segmenter
+    /// Text-promptable video segmentation (SAM 3 video). Separate from `segmenter`
+    /// because the bundle carries a `runtime` block with the memory-bank geometry.
```
alejandro-isaza: superfluous explanation, if anything rename `segmenter` to `imageSegmenter`

swift/Sources/CoreAIShared/Runtime/ModelStructure.swift
```diff
@@ -76,6 +90,9 @@ public enum ModelStructure: Equatable, Sendable, CustomStringConvertible {
             var opts = SpecializationOptions(preferredComputeUnitKind: .gpu)
             opts.expectFrequentReshapes = true
             return opts
+        case .videoSegmenter:
+            // TODO: the optimized export will need different specialization options.
```
alejandro-isaza: Resolve this to do.

swift/Sources/CoreAIVideoSegmenter/Session/MemoryBankPacker.swift
```diff
@@ -0,0 +1,332 @@
...
+/// Selects each object's eligible memories and packs them into the graph's fixed slots.
+///
+/// HF builds the tracker's memory with a variable-length `torch.cat`. A traced graph takes
+/// one shape, so this uses a fixed bank plus an additive key mask. Slot order carries no
+/// meaning, since cross-attention is permutation-invariant over keys.
+///
+/// Only the layout diverges from upstream. `gatherMemoryFrames` and `objectPointers` port
+/// `_gather_memory_frame_outputs` and `_get_object_pointers`.
+@VideoSegmentationActor
+final class MemoryBankPacker {
```
carinapeng, line 32: Interesting design on the idea of memory bank packer! Pytorch attends to all past memory and on device we attend to a fixed window, are we trading accuracy here? Does this mean parity w the pytorch reference is inherently bounded by the slot budget? Ie. when memories overflow, we drop what the reference keeps?

The idea here is we may want to make some of that divergence observable - per discussion w @guru-desh on some of the parity efforts in general we should be pursuing w torch

swift/Sources/CoreAIVideoSegmenter/Session/VideoInferenceSession.swift
```diff
@@ -0,0 +1,260 @@
...
+            nonConditioning[frame]?.memoryPositionEncoding = positionEncoding
+        }
+    }
+
+    /// Drop history that can no longer be read.
+    ///
+    /// Upstream retains everything, which a long video outgrows. The window covers what the
+    /// two readers can still reach. Conditioning entries are kept whole. Pointers iterate all
+    /// of them and there are few.
+    mutating func prune(before frame: Int, memoryWindow: Int, pointerWindow: Int) {
```
carinapeng: We have the prune mechanism but similar to kv cache eviction we can't just rely on the idea that we keep something around bc it's small, it doesn't scale to tens of thousands of frames

Does this ensure / do we have a mechanism to ensure that everything retained per frame have an eviction story like the memory bank does?

swift/Sources/CoreAIVideoSegmenter/Tracking/FrameProcessor.swift
```diff
@@ -0,0 +1,434 @@
...
+            session: session, frameIndex: frameIndex, reverse: reverse,
+            detections: detections, trackerLogits: &trackerLogits,
+            trackerScoreLogits: trackerScoreLogits)
+
+        // Execute: seed new objects, drop removed ones.
+        try await execute(
+            session: session, frameIndex: frameIndex, totalFrames: totalFrames,
+            reverse: reverse, detections: detections, plan: planning.plan)
+
+        let output = buildOutputs(
```
carinapeng, line 100: Output masks/scores zipped against the post-removal object list here, if one list is snapshotted before a deletion and the one after, this is a classic iterate two collections in lockstep but one changed underneath type of bug

buildOutputs reads session.objectIDs after execute() removed objects, then zips it with trackerLogits/trackerScoreLogits that were built over the pre-execution order

swift/Sources/CoreAIVideoSegmenter/Session/MemoryBankPacker.swift
kevchengcodes: This is a good question! We provide a user-defined budget for the memory bank to allow for a statically shaped export. Since this affects the tracker module per frame, I'd want to keep it static for this first iteration to avoid any re-specialization costs. The budget can be easily changed during export time via `--spatial-slots` (default = PyTorch as of now) and `--ptr-slots`.

models/sam3_video/README.md
kevchengcodes: I renamed it to `--input-video`, and added #260 to follow-up on `image-segmenter` CLI.

kevchengcodes: This was over-explaining an implementation detail, so I removed it.

swift/Sources/CoreAIVideoSegmenter/Session/MemoryBankPacker.swift
kevchengcodes: I thought about this for a bit, and renamed the "valid" vars to `spatialSlotOccupancy` and `pointerSlotOccupancy`, since they indicate whether that slot is actually occupied or if it's ignorable padding.

swift/Sources/CoreAIVideoSegmenter/Session/VideoInferenceSession.swift
kevchengcodes: This is a good catch, it was following the upstream behavior from transformers, but we can diverge here to make sure we're not retaining unusable frames for very long videos, since we discard them from the memory bank packer anyway. This will now evict frames that are no longer usable.

swift/Sources/CoreAIVideoSegmenter/Tracking/FrameProcessor.swift
kevchengcodes: Good catch, this was also following upstream behavior. But I think it's better to diverge per your suggestion. This will now pair the objects list with the tracked object IDs post-removal so there is no mask/score misalignment.

stikves (approved):

tjia1818 (approved):

swift/Sources/CoreAIVideoSegmenter/Tracking/FrameProcessor.swift
carinapeng: Thanks for the fix! Yeah if upstream behavior is not correct it's okay to diverge

Can you please add a test that drives a frame w a mid frame removal and assert the alignment for the object id and mask?

carinapeng (approved):
Thanks for fixing the correctness bug, if we can add a test for that before merging (See the follow up comment) since it's delicate to miss it, that'd be great

swift/Sources/CoreAIVideoSegmenter/Tracking/FrameProcessor.swift
kevchengcodes: I believe this test suite `BuildOutputsPairingTests` covers that!

PR #263, opened 2026-09-20
Add reasoning_effort control to the LLM server

stikves:
Adds an OpenAI-compatible `reasoning_effort` control so callers can set the thinking budget of reasoning models through the chat completions API or a server default.

## What this adds
- `ChatCompletionRequest.reasoning_effort` (optional string). Canonical values are `none`, `low`, `medium`, `high`; model-specific values pass through.
- `ReasoningEffort`, which maps the value to chat-template keyword arguments and applies request-over-default precedence.
- The server injects the value through `applyChatTemplate(additionalContext:)`.
- `--reasoning-default <level>` sets the server-wide default. `--no-thinking` is now an alias for `--reasoning-default none`.

## Behavior
A request that omits the field with no default set is unchanged, so current behavior is preserved. A reasoning model picks up the value when its chat template reads one of `reasoning_effort`, `reasoning_strength`, or `enable_thinking`; other models keep their existing output.

## Testing
- `ReasoningEffortTests` covers the resolver mapping, precedence, and request decoding.
- Builds and tests pass via `xcodebuild -scheme coreai-models-Package`.

Opening as a draft for early review. A FoundationModels-path variant (using `ContextOptions.reasoningLevel`) can follow as a separate change.


Lewis300 (approved):

swift/Sources/Tools/llm-server/LLMServerMain.swift
```diff
@@ -68,9 +68,15 @@ struct LLMServer: AsyncParsableCommand {
     @Option(name: .customLong("max-queue-depth"), help: "Max requests queued before returning 429 (default: 16)")
     var maxQueueDepth: Int = 16
 
-    @Flag(name: .customLong("no-thinking"), help: "Disable thinking/reasoning (appends /no_think or sets template)")
+    @Flag(name: .customLong("no-thinking"), help: "Disable thinking/reasoning (alias for --reasoning-default none)")
     var noThinking: Bool = false
 
+    @Option(
+        name: .customLong("reasoning-default"),
```
tjia1818: why we are not calling it `default-reasoning-effort` instead? It's now confusing to understand and distinguish "reasoning default" and "reasoning effort"

stikves: Makes sense, let's rename

PR #265, opened 2026-09-20
Add reproducible generation: seed and system_fingerprint

stikves:
Adds reproducible generation to the LLM server: an OpenAI-compatible `seed` request field and a `system_fingerprint` response field, backed by a deterministic seeded generator.

## What this adds
- `SeededRandomNumberGenerator` (SplitMix64) and a `seed` on `SamplingConfiguration`. Each generation step derives its own generator from `(seed, step)`, so reproducibility survives prefix reuse, reset, and cancellation.
- `seed` request field and `system_fingerprint` response/stream field on the chat completions API.
- Constrained (guided) decoding honors the seed, so guided generation is reproducible too.

## Behavior
With a seed set, the same seed, prompt, and logits reproduce the same tokens on repeated runs. Without a seed, sampling uses the system generator (unchanged). This does not guarantee determinism across machines or runtime environments.

## Testing
- `SeededSamplingTests` and `ServerAPISeedTests` (16 tests) cover seeded sampling, per-step derivation, the `seed` request field, and `system_fingerprint`.
- Build and `swift format lint --strict` pass.


swift/Sources/CoreAILanguageModels/Samplers/SamplingConfiguration.swift
```diff
@@ -121,6 +121,19 @@ public struct SamplingConfiguration: Sendable, Equatable, Hashable {
...
+    /// - **nil**: The engine samples with the system generator; output is not reproducible.
+    /// - **set**: Sampling uses a seeded, deterministic generator, so the same seed, prompt, and
+    ///   logits reproduce the same tokens on repeated runs. This does not guarantee determinism
+    ///   across machines or runtime environments.
+    ///
+    /// Each generation step derives its own generator from `(seed, step)`, so reproducibility
+    /// does not depend on a running generator's stream position. It therefore survives prefix
+    /// reuse, reset, and cancellation. Greedy sampling (temperature 0) is already deterministic
+    /// and ignores this field.
+    public let seed: UInt64?
```
kevchengcodes, line 138: Could note here that on-GPU is not supported in `CoreAIPipelinedEngine`

swift/Sources/CoreAILanguageModels/Samplers/SamplingConfiguration.swift
```diff
@@ -329,8 +348,13 @@ extension SamplingConfiguration {
     /// - Parameters:
     ///   - logits: Mutable array of Float16 logits. May be modified during sampling.
     ///   - tokenHistory: Recent token IDs for repetition penalty.
+    ///   - step: Generation step index, used to derive this step's generator when `seed` is set.
     /// - Returns: The sampled token ID.
-    public func fallbackSampler(from logits: inout [LogitsScalarType], tokenHistory: some Collection<Int32>) -> Int32 {
+    public func fallbackSampler(
+        from logits: inout [LogitsScalarType],
+        tokenHistory: some Collection<Int32>,
+        step: Int = 0
```
srjoglekar246: Won't clients see weird behavior if they don't know about `step` and this default value keeps getting applied for every call? Should we make it required?

stikves: This is to keep backwards compatibility.

It would not affect existing callers as they don't set `seed`. And those who set seed, ... shall we make this required w/ seed?

That might be the best compromise.

srjoglekar246 (approved):

swift/Sources/CoreAILanguageModels/Samplers/SamplingConfiguration.swift
```diff
@@ -340,6 +369,31 @@ extension SamplingConfiguration {
...
+    ///   - logits: Mutable array of Float16 logits. May be modified during sampling.
+    ///   - step: Generation step index. Required when `seed` is set (each step derives its own
+    ///     generator from `(seed, step)`); ignored otherwise.
+    /// - Returns: The sampled token ID.
+    public func sampleToken(from logits: inout [LogitsScalarType], step: Int?) -> Int32 {
+        guard let seed else {
+            return CompositeSampler.sample(from: &logits, config: self)
+        }
+        guard let step else {
+            preconditionFailure("A generation step index is required when a sampling seed is set.")
```
carinapeng, line 394: This would crash the entire process. In a server one poorly constructed sampling config would abort the server rather than fail one request. Suggest a thrown error and use an `InferenceRuntimeError.invalidArgument` for instance

stikves: The engine is responsible for generating the `step` parameter, and it is a non-recoverable logic error.

It is not a parameter users can set.

stikves: (More specifically this is on the hot path, and we should not add unnecessary `throws` which would cascade to all callers)

PR #266, opened 2026-09-20
Add --replay mode to the LLM server

stikves:
Adds a `--replay` mode to the LLM server: it reads a JSONL file of chat requests, runs them through the same request-handling core as the HTTP path, and writes a JSONL of results. This enables offline, functional testing of the server without opening a port.

> Stacks on #265 (reproducible generation) — the replay result format carries `system_fingerprint` and the seeded fields. Review/merge after #265.

## What this adds
- `ReplayTypes` (shared request/result JSONL format) in `CoreAILMCommon`.
- `ReplayRunner` and a `--replay <input> --replay-output <output>` flag on the server.
- The HTTP handler's core is reused, so replayed requests exercise the same path as live ones.

## Testing
- `ReplayTypesTests` covers the JSONL round-trip.
- Build and `swift format lint --strict` pass.


swift/Sources/CoreAILMCommon/ReplayTypes.swift
```diff
@@ -0,0 +1,158 @@
...
+    /// Whether this request asks for the streaming path.
+    public var isStreaming: Bool { request.stream == true }
+}
+
+/// Per-request result with instrumentation, emitted as one JSON object per line.
+public struct ReplayResult: Encodable, Sendable {
+    public let id: String?
+    public let session: String?
+    public let t: Double?
+    public let content: String?
```
kevchengcodes, line 41: why not keep the `ChatCompletionRequest` format, where we'd have a `[ChatMessage]` instead of pulling out the content and toolCalls? Would this work if there are multiple messages for one replay result?

stikves: Makes sense

PR #268, opened 2026-09-20
Fix BFloat16 KV-state crash in reset and cache growth

stikves:
Models with BFloat16 KV states (e.g. `gpt-oss-20b`) crash on any path that resets or grows state — session isolation, prefix reuse, multi-turn, guided decoding, tool loops — with `Type Float16 does not match scalar type Optional(BFloat16)`. Plain in-capacity generation is unaffected.

`zeroFillNDArray` and `copyCache` grouped `.bfloat16` with `.float16` and took a typed `view(as: Float16.self)`, which traps on a BFloat16 array. Both are raw byte operations, so the fix reinterprets the 16-bit elements through a raw view (as `flattenBFloat16NDArray` already does). Float16/Float32 behavior is unchanged.

Adds a BFloat16 zero-fill test that traps on the old code and passes now.


Lewis300 (approved):

tjia1818 (approved):

swift/Sources/CoreAILanguageModels/Handlers/StateHandler+NDArray.swift
```diff
@@ -191,9 +195,10 @@ func zeroFillNDArray(_ array: inout NDArray) {
     let count = array.shape.reduce(1, *)
     switch array.scalarType {
     case .float16, .bfloat16:
-        var view = array.mutableView(as: Float16.self)
-        view.withUnsafeMutablePointer { ptr, _, _ in
-            memset(ptr, 0, count * MemoryLayout<Float16>.size)
+        // Both 16-bit types zero to an all-zero bit pattern, so a raw view avoids the
+        // scalar-type trap a typed `mutableView(as: Float16.self)` hits on a BFloat16 array.
+        array.mutableRawView().withUnsafeMutableBytes { ptr, _, _ in
+            memset(ptr, 0, count * MemoryLayout<UInt16>.size)
```
alejandro-isaza:
```suggestion
            memset(ptr, 0, count * MemoryLayout<UInt16>.stride)
```

alejandro-isaza (approved):

PR #269, opened 2026-09-20
Honor additional stop tokens in the VLM adapter

stikves:
The Foundation Models VLM adapter (`CoreAIVisionLanguageModel`) stopped generation only on the main EOS and a hardcoded `<|im_end|>` lookup, so a VLM whose turn terminator is anything else (Gemma's `<end_of_turn>`, Phi's `<|end|>`) never stopped and ran to the response-token cap.

The text server already resolves these via `LanguageConfig.additionalStopTokenIds`. This reads the same set once at load and unions it into the stop set. Models that use `<|im_end|>` are unaffected.


swift/Sources/CoreAILanguageModels/ToolCallParser.swift
```diff
@@ -10,7 +10,7 @@ import Tokenizers
 
 extension Tokenizer {
     /// Whether `token` is a genuine entry in the vocabulary, not an unk-token fallback.
-    func vocabContains(_ token: String) -> Bool {
+    public func vocabContains(_ token: String) -> Bool {
         guard let id = convertTokenToId(token) else { return false }
         return convertIdToToken(id) == token
     }
```
kevchengcodes: should this be moved to a tokenizer file instead of tool call parser?

stikves: Tokenizer is not ours... Thinking back... this should not be public. Will make it internal and not share.

